PR #1, opened 2016-03-18
Support for private columns and filtered objects

svenbw:
When one wants to send an ORM object trough an API some columns should be remain private or hidden. This can be done using the `$_private` variable in the model.
Calls like `as_array()` will return only the non private column values.
An extra function `as_object()` is also added this will return a new object instance with the column values, it differs with the casted array (`(object)$some_model->as_array()`) that each value will be returned as it's native type (`string`, `int` or `float`).

If the functions are called with the TRUE argument (`as_array(TRUE) ` or `as_object(TRUE)`) *all* column values will be returned.
Also leaving `$_private` empty will return all column values which assures backwards compatibility.

### DIFF
--- a/modules/orm/classes/Kohana/ORM.php
+++ b/modules/orm/classes/Kohana/ORM.php
@@ -245,6 +245,12 @@ class Kohana_ORM extends Model implements serializable {
 	protected $_errors_filename = NULL;
 
 	/**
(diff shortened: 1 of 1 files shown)


neo22s:
Can we please create an issue ?

Looks a good idea to me but lets discuss it a bit more, also we should have test cases for this ;)

PR #9, opened 2016-09-06
Implements behavior support for ORM

svenbw:
behaviors allows adding custom columns automatically.
Sometimes one wants to add a guid column to a model to use in urls which are visible to the user, since there are more models that can include this type of column one would create a subclass of ORM which implements this feature.
Typically one would implement a subclass.

Behaviors allow you to add more than one custom columns which are managed in background and which can be used to query models...

In the line `$project = ORM::factory('Project', $id);` $id can be a number (real primary key) or a guid if the guid behavior is added to the class.

Behaviors are added just like rules and filters:
```
public function behaviors()
{
  return array(
    'guid' => array(),
    'slug' => array('source'=>'name'),
  );
}
```
By default guid and slug are available.

For each behavior a separate class can be implemented.

### DIFF
--- a/modules/orm/classes/Kohana/ORM.php
+++ b/modules/orm/classes/Kohana/ORM.php
@@ -245,6 +245,12 @@ class Kohana_ORM extends Model implements serializable {
 	protected $_errors_filename = NULL;
 
 	/**
+	 * List of behaviors
+	 * @var array
+	 */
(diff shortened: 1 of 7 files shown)


neo22s:
looks good! but koseven is dead... I am moving to laravel :(

PR #15, opened 2017-01-11
ORM changes

svenbw:
Corrected return values for `get_typed` function.
Added a (static) helper function `quote_table` that allows quoting the table of used in the model (based on the name). This is handy when writing "native" queries that use tables from models.
Function create will also load the insert id as the primary key if the primary key is NULL.
Furthermore I fixed some indents of a previous change.

### DIFF
--- a/modules/orm/classes/Kohana/ORM.php
+++ b/modules/orm/classes/Kohana/ORM.php
@@ -856,6 +856,10 @@ class Kohana_ORM extends Model implements serializable {
 	{
 		$value = $this->get($column);
 		
+		if ($value === NULL)
+			return NULL;
+
+		// Call __get for any user processing
 		switch($this->table_column_type($column))
 		{
 			case 'float':  return floatval($this->__get($column));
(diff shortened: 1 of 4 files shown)


svenbw:
When there is a has_many relation in an ORM model one needs to add and delete relations one by one... I modified the `set` function so it will do this by itself.

Since it might break things there is the requirement to add another array key "`update`" (that needs to be set to TRUE) to run the code. This keeps things backward compatible.

A relation definition for a **page** model will look like the code below, '`tags`' is a list of all tags, '`page_templatetags`' the link table. '`update => TRUE`' will trigger the magic.

`'tags'  => array('model' => 'Tag', 'through' => 'page_templatetags', 'update' => TRUE)`

neo22s:
we need to document this and see if breaks anything before merging it ;)

svenbw:
Yes, I always try to document the usage in the PR, and add a header for each function so the documents remain up to date (if they are generated from the source code).

svenbw:
Documentation for behaviors and support for has_many relations added to ORM guid

PR #21, opened 2017-01-17
Reintegration of collapsing menus.

Hyaena:


neo22s:
What is this change doing?

Hyaena:
It collapses the menu tree in userguide/API browser. Using the + signs you can open it and navigate. It’s been there before, so I reintegrated it. To my mind it is a lot better with collapsing menus. Attached screenshots show the difference.


after
before

PR #31, opened 2017-01-27
Default type "MySQLi"

stik:
default type changed to MySQLi


neo22s:
Thanks!! didnt realize ;)

Issue #30, opened 2017-01-27
Development flow


neo22s:
Should we commit everything into Master, and then just do a Tag per release? Is what I had in mind now.

I want to keep things as simple as possible so its easy to collaborate.

mercury64:
I would suggest create a 'develop' branch, everything commit into develop, then merge to master and tag. I never commit directly into master. I suggest a simplified version of this http://nvie.com/posts/a-successful-git-branching-model/

neo22s:
ok! sounds good to me to. 

To keep it simple and please let me know if you all agree:

1. Commit all to devel
2. Merge to Master from Devel once tested in Devel (normally before release?)
3. Tag for release

Master should be latest stable? so same as the last tag?

If we agree I will add this simple flow to the docs.

svenbw:
Yes. Agreed.

jstrobel:
+1

neo22s:
ok, new guides about how to contribute
https://github.com/koseven/koseven/blob/devel/CONTRIBUTING.md

New branch
https://github.com/koseven/koseven/tree/devel

Now devel is the default branch too
https://github.com/koseven/koseven

We will not accept PR to Master ;)

PR #35, opened 2017-02-01
Encrypt as module: solution for deprecated Mcrypt in PHP 7.1

### LINKED ISSUE #14: PHP 7.1: Deprecated Mcrypt
https://github.com/kohana/core/issues/686
https://github.com/yclas/yclas/issues/2176

svenbw:
Should fix #14.
Encrypt class is moved to a module, there are two options: Original Mcrypt and a replacement based on openssl. It is possible to add additional encryption engines without breaking the old ones.


neo22s:
damn! this is great job!! please let me review carefully, thanks a lot

svenbw:
Thanks. I needed to take some compromises to keep things backward compatible.

modules/encrypt/classes/Kohana/Encrypt.php
```diff
@@ -0,0 +1,147 @@
...
+			{
+				// No default encryption key is provided!
+				throw new Kohana_Exception('No encryption key is defined in the encryption configuration group: :group',
+					array(':group' => $name));
+			}
+
+			if ( ! isset($config['mode']))
+			{
+				// Add the default mode
+				$config['mode'] = MCRYPT_MODE_NOFB;
```
piotrgolasz: Shouldn't the default mode be changed to openssl?

modules/encrypt/classes/Kohana/Encrypt.php
```diff
@@ -0,0 +1,147 @@
...
+			}
+
+			if ( ! isset($config['cipher']))
+			{
+				// Add the default cipher
+				$config['cipher'] = MCRYPT_RIJNDAEL_128;
```
piotrgolasz: shouldn't the cipher type be changed to aes-256-ctr (or aes-256-cbc) for that matter?

modules/encrypt/classes/Kohana/Encrypt.php
```diff
@@ -0,0 +1,147 @@
...
+	 *
+	 * @param   string  $key_config    encryption key or config array
+	 * @param   string  $mode          encryption mode
+	 * @param   string  $cipher        encryption cipher
+	 */
+	public function __construct($key_config, $mode = NULL, $cipher = NULL)
+	{
+		if (is_string($key_config))
+		{
+			$this->_engine = new Encrypt_Engine_Mcrypt($key_config, $mode, $cipher);
```
piotrgolasz: The default should be Openssl imho.

modules/encrypt/classes/Kohana/Encrypt/Engine/Mcrypt.php
```diff
@@ -154,11 +64,8 @@ public function __construct($key, $mode, $cipher)
 	 * @param   string  $data   data to be encrypted
 	 * @return  string
 	 */
-	public function encode($data)
+	public function encrypt($data, $iv)
```
piotrgolasz: We should always assume that a programmer programming is a monkey, with no knowledge of encryption. I would therefore suggest not including the IV in encrypt() function, but instead automatically generate the most random IV.

modules/encrypt/config/encrypt.php
```diff
@@ -3,6 +3,7 @@
 return array(
 
 	'default' => array(
+		'type'       => 'mcrypt',
```
piotrgolasz: Use openssl as default, pretty please?

modules/encrypt/classes/Kohana/Encrypt.php
svenbw: I wanted to keep the code compatible with the original implementation: This assures that updating the Kohana system doesn't break old code or starts encrypting stuff using another method.

modules/encrypt/classes/Kohana/Encrypt.php
svenbw: Again... compatibility

modules/encrypt/config/encrypt.php
svenbw: I think this van be changed to the new default, existing installations have an existing config.

svenbw:
@piotrgolasz you're seem to know what you're doing and provide solid feedback. Do you think these changes make the modifications acceptable?
Note that an encryption type is now required, using this module in existing configurations will require a change in the code (pass a configuration array).
I still need to fix the iv thing.

neo22s:
@svenbw we should do something so the code doesnt need to change... maybe a hack.. like hardcoded config? the rest looks great to me. I haven't tested it need to be hones, only read the code.

piotrgolasz:
@svenbw https://github.com/piotrgolasz/kohana-encrypt I created my own solution for personal use. On the other topic you could create something like legacy_decrypt() where it would decrypt with mcrypt and encrypt with openssl for compatibility

svenbw:
@piotrgolasz as @neo22s mentioned it's preferable that the code remains compatible the existing code.
I see your point, but for the time being I think we should remain compatible with the existing implementation.

svenbw:
PR on development branch under #39, this one is on the master. Following the development flow (#30) this can be closed.

Issue #36, opened 2017-02-04
Can we name and add this to Packagist?


Azuka:
I'm currently using the Kohana 3.4 packages from composer on a site I hope to upgrade to PHP 7 to take advantage of the performance improvements.

Would it make sense to get this added to packagist, at least for easy updates? Although it's a full framework, it should be relatively trivial to continue to use an existing Kohana site with modules and the system directory pointing to the appropriate locations in `vendor`. `koseven/koseven` or `koseven/framework` would be a good name (pending the decision to split into separate modules and core).

I can put in a pull request for a change to the composer.json file if everyone thinks it's a good idea.

Thoughts?

neo22s:
good for me!

Issue #33, opened 2017-01-30
PHP 5 Compatibility


neo22s:
Should we maintain the software compatible? PHP 5.6 will have support until 31 December 2018. Support, so only bug fixing.

Not sure what you all think, your input is important.

What I am sure:  I will not test on PHP 5 at all. xD Someone wants to take care of it?

Gazoo:
I think with the recent news that Kohana is completely shutting down that maintaining PHP 5 compatibility is might be important (at least in the short term).

There are a lot of Kohana projects that are going to be turning to Koseven now.

neo22s:
ok, agreed we will try to keep it compatible with PHP 5 an 7 as much as we can. Once https://github.com/koseven/koseven/issues/29 is closed will be easier. thanks!

PR #37, opened 2017-02-05
Add composer definition and .gitignore.

Azuka:
From the suggestion on #36

Comments below.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,48 @@
 {
-	"require": {
-		"phpunit/phpunit": "3.7.24 - 4",
-		"phing/phing": "dev-master"
-	}
+  "name": "koseven/koseven",
+  "description": "Koseven Framework - PHP-7 compatible replacement for Kohana",
+  "license": "BSD",
+  "keywords": [
+    "kohana",
+    "kohana-php7",
+    "koseven"
+  ],
+  "homepage": "https://koseven.ga/",
+  "authors": [
+    {
+      "name": "Koseven Team",
+      "homepage": "https://koseven.ga/",
+      "role": "developer"
+    },
+    {
+      "name": "Kohana Team",
+      "email": "team@kohanaframework.org",
+      "homepage": "http://kohanaframework.org/team",
+      "role": "developer"
+    }
+  ],
+  "autoload": {
+    "psr-0": {
+      "": [
+        "system/classes"
+      ]
+    }
+  },
+  "support": {
+    "forum": "https://telegram.me/koseven",
+    "issues": "https://github.com/koseven/koseven/issues"
+  },
+  "require": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,6 +1,48 @@
...
+      "name": "Kohana Team",
+      "email": "team@kohanaframework.org",
+      "homepage": "http://kohanaframework.org/team",
+      "role": "developer"
+    }
+  ],
+  "autoload": {
+    "psr-0": {
+      "": [
+        "system/classes"
```
Azuka: I added only system based on my usage locally. Others can be added as needed to project composer.json files.

composer.json
```diff
@@ -1,6 +1,48 @@
...
+  },
+  "require-dev": {
+    "phpunit/phpunit": "^6.0",
+    "phing/phing": "^2.16"
+  },
+  "conflict": {
+    "kohana/core": "3.3.*"
+  },
+  "replace": {
+    "kohana/core": "3.3.*"
```
Azuka: This and conflict are important to prevent other packages that rely on kohana core from installing the legacy version e.g. I use [alle/kohana-jsend](https://github.com/alle/kohana-jsend)

neo22s:
looks good to me

Azuka:
Cool, want to submit this to packagist, or should I?

neo22s:
go ahead! no problem ;) thanks

Azuka:
We need a [service hook](https://packagist.org/about#how-to-update-packages) to get it to auto-update.

neo22s:
I need this 

image

You want to send it to me privately on telegram?

thanks

svenbw:
First of all: good job! Two remarks:
Should we finally drop support for older PHP versions? I think this version still works on older versions.
The kahanaframework.org URL will go down 1ste july

neo22s:
hook done, how can we test it?

Azuka:
@svenbw is this supposed to work with both PHP 5 and 7?

@neo22s there should be a test service button.

neo22s:
I clicked on test service, not sure what to expect sorry.

@svenbw @Azuka for me I will drop compatibility to PHP 5 but is something to decide all of us.

Azuka:
@neo22s, I see on my end that the hook works.

We may want to open a new issue to let others vote, but if we want to keep PHP 5 compatibility, we'd need to remove that line from the composer.json file.

neo22s:
@Azuka already did https://github.com/koseven/koseven/issues/33 no one answered hahaha

svenbw:
In my opinion: It's supported by many hosting providers as one of the two choices (5.x and 7.x), it works with old modules and makes it possible to update Kohana to Koseven. But as you say: let's decide together.

PR #62, opened 2017-03-08
Fix #59: pagination controller and action.

### LINKED ISSUE #59: Pagination - Required route parameter not passed: action
Hi there,

I use the Pagination module, which is now included in this repo.

Right now I'm getting this error:

`EMERGENCY: Kohana_Exception [ 0 ]: Required route parameter not passed: action ~ SYSPATH/classes/Kohana/Route.php [ 580 ]`

I tried to debug that and the details are:

In the pagination/classes/Kohana/Pagination.php there is **url()** method... and inside there is this part of code:

```
case 'query_string':

	return URL::site($this->_route->uri($this->_route_params).
		$this->query(array($this->config['current_page']['key'] => $page)));
```

The **$this->_route->uri** expects the array as a parametr... and one of the keys of this array should be "action" key.

**But it isn't.**

The "action" key is unset in the Request class on purpos (because there is public $request->action() method), so I beleive there is some bug in the Pagination module.

If I hotfix the Pagination contructor like this:

```
// Assign default route params
$this->_route_params = $request->param();
$this->_route_params['action'] = $request->action();  // <-- this is my hotfix
```

it works just fine without any error.

Could you please verify my statement?

Azuka:
Adds the params for controller and action to the route params in pagination for routes with variable params.

### DIFF
--- a/modules/pagination/classes/kohana/pagination.php
+++ b/modules/pagination/classes/kohana/pagination.php
@@ -97,7 +97,11 @@ class Kohana_Pagination {
 		
 		// Assign default route params
 		$this->_route_params = $request->param();
-		
+
+		// Add controller and action to route params for routes with variable controllers and actions
+		$this->_route_params['controller'] = $request->controller();
+		$this->_route_params['action'] = $request->action();
+
 		// Pagination setup
 		$this->setup($config);
 	}


neo22s:
I tested in local in a testing site for Yclas where I have different places pagination setup, using different routes etc. Seems to be fine, I never had an issue anyway but at least does not break anything ;)

PR #70, opened 2017-03-24
Dockblock

adamcopley:
Have updated copyright years and url for new koseven license. The license needs uploading to the URL before this is merged in as it currently shows a 404 error. https://koseven.ga/LICENSE.md


neo22s:
maybe we should remove the dates... and thats all hehehe

also thers a conflict someone can take a look? thanks

PR #73, opened 2017-03-24
Merge of Encrypt module branch

svenbw:
Support both mcrypt and openssl encryption methods trough a encrypt module.
Solves #14


neo22s:
Theres an error on Travis ;)

piotrgolasz:
Error says "Class Encrypt not found". This module has to be enabled in bootstrap ;)

svenbw:
I've loaded the module by modifying the test bootstrap file (loading the module there, just like the unittest module). The second option is to enable it by default in the Kohana bootstrap.php file.

Unit tests pass, but not for PHP7 since the default test use mcrypt in the tests.

neo22s:
this looks good to me! someone else could test it?

I will just merge it and test better , thanks a lot @svenbw

PR #74, opened 2017-03-27
Memcached

adamcopley:
Add kohana-memcached files, sanitize id's in all get & set methods, add config.memcached.php


modules/cache/classes/Kohana/Cache/Memcached.php
```diff
@@ -0,0 +1,188 @@
+<?php defined('SYSPATH') or die('No direct script access.');
```
neo22s: we have removed `defined('SYSPATH') or die('No direct script access.');` from all files ;)

remove it and we are fine! :D

PR #80, opened 2017-04-04
Replaced download link

knbv:
Users should see a list of changes ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 [![Telegram](https://img.shields.io/badge/Telegram-joinChat-blue.svg)](https://telegram.me/koseven)
 [![Build Status](https://travis-ci.org/koseven/koseven.svg?branch=devel)](https://travis-ci.org/koseven/koseven)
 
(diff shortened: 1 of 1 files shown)


neo22s:
but also the download link I think...

knbv:
There are also links for downloading
https://github.com/koseven/koseven/releases/tag/3.3.7

neo22s:
direct link I mean hehehe but doesnt matter

neo22s:
here should be done also https://github.com/koseven/koseven/blob/gh-pages/index.md

And please before submitting PR we need a issue so we can discuss before we do the work. To avoid this. I like everything to be voted, I feel bad about closing a PR without merging it.

knbv:
Sorry, I thought that this is too small a change to create a discussion :)

Thank you!

knbv:
Sorry, I thought that this is too small a change to create a discussion. However, you can discuss it right here :)

Thank you!

neo22s:
no problem heheh its to keep track of changes too. thanks a lot.

PR #108, opened 2017-06-17
Module's Minion and Image - reduction to Kohana-codestyle

arkagod:
Module's Minion and Image - reduction to Kohana-codestyle


mzjmc123:
Kohana why can't you find the official documents?

PR #112, opened 2017-06-22
Correct expected keys using ORM values method

rafi:
Somewhere along the line `ORM::values` got a new required argument, expected keys, but `Kohana_Auth_ORM` wasn't adapted.


svenbw:
If you look at [the values implementation](https://github.com/koseven/koseven/blob/devel/modules/orm/classes/Kohana/ORM.php#L835), you'll see that the `$expected` argument is still optional.

rafi:
My bad, I was looking at a version of the ORM which had it required, i.e. `values(array $values, array $columns)`

svenbw:
No problem!

PR #113, opened 2017-06-22
update folder structure for ease of deployment

Samnan:
I have updated the folder structure and tested that install.php works out of the box, and it is now also easy to keep the framework related files anywhere, by just changing one line on top on index.php file. This would facilitate kohana deployment on different hosting providers.


Samnan:
Ok passed travis checks now.

Not sure which part does not follow kohana code style? its a one line change preceded by comments :)

knbv:
> Not sure which part does not follow kohana code style?

The problem with concatenating strings in the file public / index.php, for example here:
https://github.com/koseven/koseven/pull/113/files#diff-8fda5307a067372337ec4b9adb16b55fL9

Documentation: https://docs.koseven.ga/guide/kohana/conventions#string-concatenation

Samnan:
ok updated kohana code style  (sorry been too long since i took a look at that document).

Samnan:
Finally, I have added .htaccess rule to deny access to the framework folder, if the following url is invoked:

'http://localhost/kohana/framework'

and assuming the framework folder is sitting besides the index.php file, either a 404 page by kohana will be shown, or if you have created a controller by the name of 'Framework', then it will be used.

piotrgolasz:
1) Where is the PR
2) I would be strongly against, as it would require to change structure in all projects while updating. I would recommend rejecting this.

neo22s:
I am the same here, it requires a deep change in all existing projects. :(

anyone else?

thanks

jstrobel:
I would be rejecting this too - because of the structure change.

Samnan:
Well i am truly disappointed. The reason i created the pull request was due
to user concerns over deployment over folder structure for different
hostings (me included).

Secondly, the change may seem large, but it is only a change in index.php
file and moving three folders in the file system slightly. For any existing
project, it is a two minute change, and all new projects would ensure
maximum portability between shared and dedicated hostings.

I remember when kohana 3.3 was released and they renamed all files and
folders for PSR0 compliance, there was a lot of noise in the community, but
eventually everyone was forced to go through the change if they wanted to
upgrade. I myself spent a week upgrading and fixing my projects. This
change is fas less dramatic and does not need any testing of whole site, if
the site works after moving the folders and updating index.php, the rest
should work as usual.

Oh well, i will stick to my fork, so long.

neo22s:
We can check this for 3.4.X but not for 3.3.X

I am sorry you are disappointed, but we can not fill all the criteria for everyone without affecting others.

Sure no problem its an open source project you can fork and have as much fun as you want, its exactly what I did.

thanks.

PR #132, opened 2017-08-04
devel

### LINKED ISSUE #131: PHP 7.2 deprecated functions
Lets make sure that koseven is PHP 7.2 compatible. I noticed that its still using "each" in system\classes\Kohana\Response.php (line 301) which has been deprecated. We can just convert this into a foreach loop.

https://wiki.php.net/rfc/deprecations_php_7_2

Gazoo:
Fixes #131

### DIFF
--- a/system/classes/Kohana/Response.php
+++ b/system/classes/Kohana/Response.php
@@ -298,10 +298,10 @@ class Kohana_Response implements HTTP_Response {
 		if (is_array($key))
 		{
 			reset($key);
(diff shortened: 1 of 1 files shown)


neo22s:
hey guys for me is looking good @jstrobel @svenbw @piotrgolasz want to take a look?

jstrobel:
Looks good for me. Just want to add the depracted hint in the german documenation, but it is already added but not published ;)

PR #139, opened 2017-08-10
Updated install guide and install image

satanch:
Updated PHP version, extensions and links.


satanch:
Updated Kodoc template and some stuff. Please check.

PR #152, opened 2017-09-06
New install page

satanch:
Completely new page without PHP 5.3 checks.

Error screenshot

Success screenshot


neo22s:
now I see we do not require mcrypt anymore since its deprecated for 7.2, maybe just remove it? since its optional. You can use openssl https://github.com/koseven/koseven/pull/73

PR #155, opened 2017-09-06
skip mcrypt unit tests if the extension is not loaded

errotan:


errotan:
Damn the unit tests fail with this. On my test virtual machine phpunit skipped them. Looking into a fix.

PR #157, opened 2017-09-08
use language order from the client's Accept-Language: HTTP header

errotan:


piotrgolasz:
Some unit tests to that?

errotan:
Added the unit tests.

system/classes/Kohana/HTTP/Header.php
```diff
@@ -811,6 +816,81 @@ public function accepts_language_at_quality($language, $explicit = FALSE)
...
+			$this->_accept_language_list = HTTP_Header::_parse_language_header_as_list($language_header);
+		}
+
+		$new_order = [];
+
+		foreach ($this->_accept_language_list as $accept_language)
+		{
+			foreach ($languages as $key => $language)
+			{
+				if (($explicit AND $accept_language == $language) OR
```
knbv: Is it possible to use a non-strict comparison here?

neo22s: then its ok or not?

knbv: I think that you need to use a strict comparison in the following lines: 875, 876

The fact is that a non-strict comparison can be unsafe. Or is it not important in this case?

PR #159, opened 2017-09-22
Openssl - as default cipher. Disable mcrypt iv in Engine_Openssl

piotrgolasz:
* Configs with all parameters
* Validating cipher key exact length disabled, minimal length added
* Removing mcrypt_get_iv from Engine_Openssl
* Engine_Openssl is now a default engine


piotrgolasz:
DO NOT MERGE - ADDITIONAL TEST HAS TO BE WRITTEN

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -44,33 +53,43 @@ public function __construct($key_config, $mode = NULL, $cipher = NULL)
...
+		
+		//Require at least 16bit key
+		if ($length < 16)
 		{
-			if ($length !== 16)
-			{
-				// No valid encryption key is provided!
-				throw new Kohana_Exception('No valid encryption key is defined in the encryption configuration: length should be 16 for AES-128-CBC');
-			}
+			return false;
```
svenbw: Use uppercase FALSE (and TRUE)?

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -44,33 +53,43 @@ public function __construct($key_config, $mode = NULL, $cipher = NULL)
...
 		}
-			
-		elseif ($this->_cipher === 'AES-256-CBC')
+		
+		$type = explode('-', strtoupper($this->_cipher));
+		
+		//Do not allow ciphers other then AES
+		if($type[0] !== self::AES)
```
svenbw: This is a strict comparison: For type $type[2] not.

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -44,33 +53,43 @@ public function __construct($key_config, $mode = NULL, $cipher = NULL)
...
+			// No valid encryption cipher is provided!
+			throw new Kohana_Exception('No valid encryption cipher is defined in the encryption configuration. Pick one of AES-***-*** ciphers. ECB type is weak and should not be used.');
+		}
+	}
+	
+	/**
+	 * Check correct cipher validation
+	 * @return bool
+	 */
+	protected function validate_cipher_type():bool
```
svenbw: Specifying the return type of a function is one of the good things of PHP7, but it's not available in other classes in the Kohana core. Maybe we should do this in one of the next milestones and leave it out here. Or should we migrate slowly?

piotrgolasz: Well, for one thing Ko7 should be PHP7 compatible. However all coding should be done according to one standard.

errotan:
Do you need help on this?

PR #161, opened 2017-09-25
Behavior fixes

svenbw:
Checking the GUID before inserting them into the database has a big performance penalty when adding a lot of data.
The new code doesn't check the GUID before inserting the row. This can be switched on by adding the `verify` flag with value `FALSE` in the behavior configuration function in the model.

Also removed the CRLF.


neo22s:
Whats the issue related? I do not understand it. Also seems theres issues when merging.

svenbw:
There is no issue logged. Everything works but there is a performance penality. Before inserting a record the GUID was checked to see if it was already in the DB. This requires some time, which results in a large performance penality. This PR fixes that.

The merge issue seems to be related to the switch from CRLF to LF.

neo22s:
Can we please create an issue first so we discuss it together before doing the PR? it's how we do all of them. I am not comfortable changing a behavior, maybe someone relies on this specific and then is breaking his app.

I can not compare code either directly from here.

Issue #111, opened 2017-06-22
Client IP recorded incorrectly when client's behind a proxy


blckct:
The code here:
https://github.com/koseven/koseven/blob/master/system/classes/Kohana/Request.php#L109

It returns the first IP in the list and so it may be a privat subnet IP. Additionally it's easy to spoof it so it may be a security risk.

neo22s:
thanks for reporting! please PR a change ;)

jstrobel:
@blckct Just to give you a hint if you don't know how: https://help.github.com/articles/creating-a-pull-request/

blckct:
@jstrobel It's a bit more complicated then that. It's more of an architecture decision rather than anything else, changing how XFF is interpreted may break a lot of things.

jstrobel:
So i think, if you share some of your ideas we can discuss it in group

blckct:
Well, the current code looks like this:

>				// Format: "X-Forwarded-For: client1, proxy1, proxy2"
> 				$client_ips = explode(',', $_SERVER['HTTP_X_FORWARDED_FOR']);  
> 				Request::$client_ip = array_shift($client_ips);

It returns the leftmost IP supplied by the user, I suggest we should use the right most that stays after removing the trusted proxies.

In fact I did a quick check and it seems that's what happens in Symfony.

satanch:
So it's fixed or not?

This is how I see our solution:
[https://repl.it/KU5e/1](https://repl.it/KU5e/1)

Is this what we want? :thinking:

neo22s:
@satanch I am not sure if its correct lets be honest. I have not tried in a real scenario where theres multiple proxies or other complex things... 

What is 100% sure is that yo can never trust the request since its easy to spoof even in this casi you could do it...

blckct:
We're already checking if the request comes from a trusted proxy so we should be pretty sure that right-most value is not spoofed if the proxy does support X-Forwarder-For

neo22s:
@blckct so please make a suggestion or a PR and we close this ;)

blckct:
I was saying @satanch solution should be working well.

satanch:
Should I create PR with my solution. I'm afraid that we can break some things and get some new bugs. :confused: 
@blckct

satanch:
Code is correct as I think.

[https://en.wikipedia.org/wiki/X-Forwarded-For](https://en.wikipedia.org/wiki/X-Forwarded-For)

> the left-most being the original client

[...]

PR #163, opened 2017-10-02
I was waiting a lot, issue #111

satanch:
So now we using rightmost IP. Okay? :thinking:


schelkun:
Install image updated too

Issue #121, opened 2017-07-05
Defining your preferred language


knbv:
To determine the most preferred language, [this function](https://github.com/koseven/koseven/blob/devel/system/classes/Kohana/HTTP/Header.php#L827) is used, however it does not work correctly in this case:

For the test is used smartphone Samsung - Galaxy A5 with Russian language enabled, PHP 5.6.30

Test 1:
`HTTP::request_headers()->preferred_language(['en', 'ru',]);`

Test 2:
`locale_accept_from_http($_SERVER['HTTP_ACCEPT_LANGUAGE'])`

![screenshot 2017-07-05 003](https://user-images.githubusercontent.com/16259603/27873916-62e28188-61b6-11e7-882c-f41891b1f1bf.png)

As you can see, tests show that the language was incorrectly defined in both cases, but why? Samsung devices send the header in the wrong format? Even if it is, we need to fix this, given that the built-in PHP function also defines the wrong language?

neo22s:
can you do a PR with a fix? thanks

errotan:
I think the problem is that @knbv 's phone sends the languages with the same 'priority'. For example using this test controller action:

`$this->response->body(HTTP::request_headers()->preferred_language(['en', 'ru', 'hu']));`

Most browsers send the above like header. The 'en' has 0.3 and 'hu' has 0.8 which makes 'hu' greater and that is displayed.

Accept-Language: hu-HU,hu;q=0.8,en-US;q=0.5,en;q=0.3

Using the above the displayed value is 'en'

Accept-Language: hu-HU,hu,en-US,en

I think this is a correct behaviour because the script can choose any, but can do a PR to use the order specified by the browser. Should I do it?

satanch:
According to RFC 2616

>    The language quality factor assigned to a language-tag by the
>    Accept-Language field is the quality value of the longest language-
>    range in the field that matches the language-tag. If no language-
>    range in the field matches the tag, the language quality factor
>    assigned is 0. If no Accept-Language header is present in the
>    request, the server

errotan:
Made a PR for this: #151

errotan:
I need to fix the code to make the unit tests pass.

errotan:
Made a PR again which less modifies the original code and passes unit tests. Please review: #157

piotrgolasz:
@knbv Can you check the solution?

knbv:
@piotrgolasz, @errotan It is checked, now the language is determined correctly, thanks!

[...]

PR #164, opened 2017-10-02
use language order from the client's Accept-Language: HTTP header #121

errotan:


knbv:
@errotan Tell me, please, why do not you use a strict comparison in lines 875,876?

errotan:
@knbv because the calling function [ preferred_language() ] also has the $explicit argument.

knbv:
@errotan Excuse me, do not quite understand how this is connected?

errotan:
I don't understand your problem. Strict comparison is only used when $explicit is TRUE, otherwise the first 2 characters should be equal.

knbv:
The substr function returns the extracted part of the string parameter; or FALSE if an error occurs or an empty string.

If you know that both values are guaranteed to be strings, then == and === are identical, because the only difference between them is that === checks whether the types are the same, not just the effective values.

Are you sure that using a non-strict comparison in this case is safe? I did not study all the code in detail, it's just a refinement, but in any case I recommend using a strict comparison.

errotan:
Ohhh, now I get it. :) I thought you are referring about the strict comparison with $explicit, didn't come to my mind that equals (==, ===) stuff. It is safe to only use non-strict (==), beacuse all input gets through atleast one strtolower() which makes it a string a regexes (like [\w-]+) makes atleast one word character.

knbv:
@errotan Thank you! ;)

errotan:
So can this be merged? Thx.

Issue #174, opened 2017-11-22
Wrong path for DOCROOT for minion tasks


Evoluty:
Hey just a little fix to do with DOCROOT const.

The path is wrong when running minion tasks from the root directory :   
`./modules/minion/minion TaskToRun`   

The path is only right when running tasks from the module directory.
It needs to be absolute and not relative to be right everywhere.

I'll do a PR to fix this.

knbv:
I have the following directory structure:

/var/www/username/data/www/example.com/

Let's write a small test:

/application/classes/Task/Test.php

```
class Task_Test extends Minion_Task
{
    protected function _execute(array $options)
    {
        Minion_CLI::write(DOCROOT);
    }
}
```
Results:
1)
index.php: 
`define('DOCROOT', realpath('../').DIRECTORY_SEPARATOR);`

result: 
`/var/www/username/data/www/`

2)
index.php: 
`define('DOCROOT', realpath(__DIR__.'/../').DIRECTORY_SEPARATOR)`

result: 
`/var/www/username/data/www/`

Both options work incorrectly, but there is one more option:

3)
index.php: 
`define('DOCROOT', realpath(dirname(__FILE__)).DIRECTORY_SEPARATOR)`

result: 
`/var/www/username/data/www/example.com/`

Check it out for yourself

Also, to run a task from the root directory, you can use this command:
`path/to/index.php --task=Test`
or copy the file `/modules/minion` to the root directory

Evoluty:
Not really sure how and from where you are running the task.  

I just made a fresh clone, enabled minion tasks (in the application/bootstrap.php), created the same task as you in application/Task and I run with `./modules/minion/minion Test`

Here are the results that I have: 

1. With the current way `define('DOCROOT', realpath('../').DIRECTORY_SEPARATOR)` 
Result: `/Users/franck/Desktop/`

2. With my version `define('DOCROOT', realpath(__DIR__.'/../').DIRECTORY_SEPARATOR)`
Result: `/Users/franck/Desktop/koseven/`

3. With your version `define('DOCROOT', realpath(dirname(__FILE__)).DIRECTORY_SEPARATOR)`
Result: `/Users/franck/Desktop/koseven/public/`

I also tried with `php public/index.php --task=Test` from the root of the project and got the same results.

Don't know why we are getting different results.
What do you think about that ?

knbv:
Sorry, there was an error, I do not use the public folder in the project where the tests were run. I approved your corrections.

[...]

PR #175, opened 2017-11-22
Fix DOCROOT path which could be wrong when using CLI minion task

Evoluty:
Fix the issue opened [here](https://github.com/koseven/koseven/issues/174)

The path is wrong when running minion tasks from the root directory :
`./modules/minion/minion TaskToRun`

The path is only right when running tasks from the module directory.
It needs to be absolute and not relative to be right everywhere.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -54,7 +54,7 @@ error_reporting(E_ALL | E_STRICT);
  */
 
 // Set the full path to the docroot
-define('DOCROOT', realpath('../').DIRECTORY_SEPARATOR);
+define('DOCROOT', realpath(__DIR__.'/../').DIRECTORY_SEPARATOR);
 
 // Make the application relative to the docroot, for symlink'd index.php
 if ( ! is_dir($application) AND is_dir(DOCROOT.$application))


piotrgolasz:
Maybe some unittest to verify this?

Evoluty:
@piotrgolasz Yes I agree.   
But I will need some help for that because I don't really know how to test this const in a unit test ☹️.

To me, the const need to be tested 2 ways:
1. When koseven is running in usual way (to be sure it's not broken for basic usage)
2. When koseven is executing a task (what this PR fix)

piotrgolasz:
I tested this on my project and it worked. So this fixes the issue, only unittest is needed @Evoluty

Evoluty:
Thanks @piotrgolasz I will try to update the PR ASAP.

piotrgolasz:
@Evoluty I did this on my branch since I don't have access to yours, add yourself into @author and add them to the project. I have php 7.2 and got a lot warnings from deprecrated mcrypt, but this should work: https://github.com/piotrgolasz/koseven/commit/63b9163cfa225429dfb6a832205b7e05e5ca223d

piotrgolasz:
BTW I run minion from index.php
```
php index.php --task=test
```

errotan:
#194 is the same as this. Should this be closed?

PR #185, opened 2017-12-13
fix for #184 - Parameter must be an array or an object that implement…

Random90:
in PHP 7.2 count() can take only arrays and Countables as parameter.

### DIFF
--- a/modules/orm/classes/Kohana/ORM.php
+++ b/modules/orm/classes/Kohana/ORM.php
@@ -1640,7 +1640,7 @@ class Kohana_ORM extends Model implements serializable {
 		}
 		else
 		{
-			return $count === ((is_int($far_keys) || is_string($far_keys)) ? 1 : count($far_keys));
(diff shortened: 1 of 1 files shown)


MHermsen:
For me it's not working ! why

Random90:
@MHermsen im 100% sure it is working

piotrgolasz:
Solution: unittests!

Random90:
@piotrgolasz he made mistake copying it to his project in #185.

piotrgolasz:
clhzedowoaa1bgp

bato3:
If `$far_keys` is digit better use `ctype_digit($far_keys)` instead `is_int($far_keys) || is_string($far_keys)`?

Or this code: https://github.com/Random90/koseven/blob/8dc9214f203c07379e567e753454db47043399a0/modules/orm/classes/Kohana/ORM.php#L1698-L1701

Random90:
@bato3 
No.
ctype_digit: Checks if all of the characters in the provided string, text, are numerical.

If far_keys would be integer, ctype_digit would return false and use count($far_keys)

errotan:
Different solution (#240) was merged.

PR #194, opened 2018-01-21
Fix Docroot Path for Minion

piotrgolasz:


piotrgolasz:
@neo22s Why are tests not performing composer install?

knbv:
#174

system/tests/kohana/MinionTaskTest.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+class MinionTaskTest extends Unittest_TestCase
+{
+
+	/**
+	 * Tests that Minion Task Help works assuming all other tasks work aswell
+	 */
+	public function test_minion_runnable()
+	{
+		$minionResponse = Minion_Task::factory(['task' => 'help']);
```
knbv: Do I understand correctly that this test simply calls a task named "help"?

The fact is that the task will be executed regardless of whether the correct path is specified in the DOCROOT.

piotrgolasz: In unittest bootstrap i included autoloader.php (https://github.com/koseven/koseven/pull/194/files#diff-d97290f5a2964424212543942a6ede21R65)

Therefore if Minion task has executed correctly, that would mean that the docroot path is also correct.

knbv: Thanks you!

knbv:
Sorry, I did not notice that you did not fix all the errors. The code does not match the Coding Style (String Concatenation).

https://docs.koseven.ga/guide/kohana/conventions

knbv:
@piotrgolasz Sorry, I did not notice that you did not fix all the errors. The code does not match the Coding Style (String Concatenation).

https://docs.koseven.ga/guide/kohana/conventions

piotrgolasz:
@knbv understood.

system/tests/kohana/MinionTaskTest.php
errotan: $minionResponse does not conform with required coding style. It should be $minion_response.

modules/unittest/bootstrap.php
```diff
@@ -61,6 +61,8 @@
  *
  * @link http://kohanaframework.org/guide/using.configuration
  */
+// Require composer libraries
+require DOCROOT.'/vendor/autoload.php';
```
errotan: This change is not required for this patch.

.travis.yml
```diff
@@ -21,6 +21,7 @@ before_install:
 # execute any number of scripts before the test run, custom env's are available as variables
 before_script:
   - sudo chmod -R 755 modules/ application/ system/
+  - composer install
```
errotan: This change is not required for this patch.

PR #195, opened 2018-02-02
Fixes #192 find_file taking long time to load files

### LINKED ISSUE #192: find_file taking long time to load files.
`find_file` function is taking almost 1sec even when cache is enabled.

I debugged the system and noticed that when `Kohana::init` is called in bootstrap, `Kohana_Core` class can't find `Kohana_Cache` file so it use file_cache function and try to load .txt file on `Kohana_Core` line **892**.

But when shutdown_handler is called and it try to write file_cache it can find `Kohana_Cache` and it use it so a cache file is created `Kohana_Cache_File` class with .cache file.

This mean when `Kohan::init` is called it never use cached file and build files list every time. 
So my solution is to update Koahan_Core class change default extension 

I try to move Kohana::modules call before Kohana::init but userguide module init.php throw errors. 

both `Kohana_Core` and `Kohana_Cache_File` are using `sha1` function to hash file name but `Kohana_Cache` use $prefix so names does not match.

So even changing file extension in  `Kohana_Cache` to .cache is not working.

Any solutions?

farazilu:
Fixes #192 
Updated the Kohan_Core::init and shut_down_handler() to use file_cache() so they can read and write to same file
Update travis.yml to use defult phpunit available as it work with php 7.0, 7.1 and 7.2


system/tests/kohana/CoreTest.php
```diff
@@ -174,6 +174,24 @@ public function test_cache($key, $value, $lifetime)
...
+	/**
+	 * Test Kohan::$caching
+	 * 
+	 * 
+	 * @test
+	 * 
+	 */
+	public function test_fild_file_work_with_caching ()
+	{
+	    kohana::$caching = TRUE;
```
piotrgolasz: Kohana with capital 'K'? what is 'kohana' class?

farazilu:
updated to capital 'K',

system/tests/kohana/CoreTest.php
```diff
@@ -174,6 +174,24 @@ public function test_cache($key, $value, $lifetime)
 		Kohana::cache($key, $value, $lifetime);
 		$this->assertEquals($value, Kohana::cache($key));
 	}
+	
+	/**
+	 * Test Kohan::$caching
```
errotan: Multiple typos in this comment and in function comments.

errotan:
One drawback of this is that if you delete an "overriden" file from the cascading file system that is cached, koseven will fail because it tries to load it. Also if a file path in the cascading filesystem is already cached, koseven will not find new files. Eg. you use Debug class and later you create a Debug.php file in application/classes/ to override. So in production where caching is probably on, you have to clear the cache folder on every new release of your app.

PR #199, opened 2018-02-26
Replaced deprecated function create_function

### LINKED ISSUE #198: Function create_function has been DEPRECATED as of PHP 7.2.0
> Given the security issues of this function (being a thin wrapper around eval()), this dated function has now been deprecated. The preferred alternative is to use anonymous functions. 

http://php.net/manual/en/migration72.deprecated.php

This function is found in several modules:

Userguide:
https://github.com/koseven/koseven/blob/devel/modules/userguide/vendor/markdown/markdown.php#L1631
Codebench:
https://github.com/koseven/koseven/blob/devel/modules/codebench/classes/Bench/AutoLinkEmails.php#L64

knbv:
Fix #198


knbv:
#202

PR #208, opened 2018-03-23
halt script execution when exception occurs in view rendering

errotan:
… status code. Please review this, I'm open to suggestions.

### DIFF
--- a/system/classes/Kohana/View.php
+++ b/system/classes/Kohana/View.php
@@ -238,14 +238,12 @@ class Kohana_View {
 		catch (Exception $e)
 		{
 			/**
-			 * Display the exception message.
+			 * Display the exception message and halt script execution.
 			 *
 			 * We use this method here because it's impossible to throw an
 			 * exception from __toString().
 			 */
-			$error_response = Kohana_Exception::_handler($e);
-
-			return $error_response->body();
+			Kohana_Exception::handler($e);
 		}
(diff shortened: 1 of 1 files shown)


errotan:
Updated.

errotan:
Currently if a view is rendered inside a view using echo or casting with (string) and there is an exception thrown inside the subview an error text is returned to the outer view but the execution continues. According to "clean code" guidelines this is "swallowing exceptions" which is bad. This patch halts the script upon exception. What do you guys think about this?

piotrgolasz:
Sounds interesting. Can you provide any example of this, for example a gist with some code in the Welcome_Controller?

errotan:
https://gist.github.com/errotan/a95c28be9b3edf636ecf05f64085b8fe

WinterSilence:
@piotrgolasz  @errotan https://github.com/koseven/koseven/blob/devel/system/guide/kohana/tutorials/basic-controller.md i add new examples in guide, but need update docs.koseven.ga.

toitzi:
Looks good.

@errotan Any updates to this? Shall this be merged?

Sorry for asking this so oftne, but i try to regulary go over all issues and PRs to see if i can help and to keep them alive

errotan:
The status is that I'm waiting for review on this. If you think it's good it can be merged.

PR #209, opened 2018-03-26
Fix warning: `current() expects parameter 1 to be array, boolean given`

juherr:


system/classes/Kohana/File.php
```diff
@@ -139,7 +139,11 @@ public static function exts_by_mime($type)
 	 */
 	public static function ext_by_mime($type)
 	{
-		return current(File::exts_by_mime($type));
+		$exts = File::exts_by_mime($type);
+		if (!$exts) {
```
errotan: Please put curly brackets to separate line according to https://docs.koseven.ga/guide/kohana/conventions

juherr: Done, thanks for the review

juherr:
Another option for the fix could be to remove `FALSE` as value of `exts_by_mime` and return `[]` but it may break things if some users are already checking against `FALSE` and not empty array.

juherr:
It was one of my options too. But as I said, it may break things because of the behavior change.
I may fix the documentation too :) Let me know what I have to do.

errotan:
I vote on clarifying the doc and to change the return type to [] in a later version,

` * @return  array|false   File extensions matching MIME type or false if none`

piotrgolasz:
I vote for this aswell. Are there any usages in Koseven of this method?

juherr:
@piotrgolasz No internal usage of exts_by_mime or ext_by_mime

juherr:
@piotrgolasz Any update on this PR?

PR #210, opened 2018-03-26
Fix error: 'Object of class stdClass could not be converted to string'

juherr:


juherr:
@errotan Done 👍

knbv:
@juherr Hmm... Perhaps you can suggest a real example where it can be needed? Why do need to pass the result of the serialize function to the quote method? How will this work? Thank you!

juherr:
@knbv I didn't digg too much and I don't know how/why the quote method is called. 

In fact, I see an issue on a project because Kohana is trying to quote a stdClass object (which is failing). 
My model contains a stdClass as attribute which is not mapped into the database.
So, I suppose Kohana is quoting not mapped attribute instead of skipping them.

BTW, I'm open to any other tips :)

piotrgolasz:
I think at this point we should also discuss using PHP 7 types....

juherr:
@piotrgolasz Let me know, but which PHP7 type is not convertible to a string by using the cast?

piotrgolasz:
I meant that original Kohana was written in php 5.x, therefore functions don't have parameter types. If Koseven should be php 7.x compatible, maybe we should think about hard typing....

juherr:
@piotrgolasz I'm not understanding why types will help here. The function is called with a mixed param and the assumption "All other objects will be converted using the `__toString` method" is not working in every case.

piotrgolasz:
@juherr Could you maybe write a unittest with code example where you pass stdClass to ORM?
Code is fairly simple, it just adds one extra case, but I can't seem to find a proper usage for this.
Your help would be greatly appreciated.

toitzi:
Maybe he used `mysql_fetch_object` somewhere or something like that. but nevermind, i am going throug unit tests this week, and unit test for this is on my todo :)

juherr:
Thanks for the merge! 👍 

The use case is still unclear for me and as I said previously:
> My model contains a stdClass as attribute which is not mapped into the database.
> So, I suppose Kohana is quoting not mapped attribute instead of skipping them.

I fact, I don't know why and when quote is used (and maybe it shouldn't in specific cases).
At least, the fix will be a workaround in some specific cases :)

PR #223, opened 2018-04-26
Update I18n.php

saruman:
If you have many modules with their own i18n folder and you want to rewrite or add new items to an specific group (array) with this little change now you can.

I have a helper module that contains most of the translations and text that I'm using in all my applications,, such as buttons, labels, menu options, etc. 

I use I18n::load('en') to get all the text and translations from all the modules that I am using in my applications, but with the array_merge doesn't include new or updated items in my applications.

That's why I'm requesting this little change to the Core.

### DIFF
--- a/system/classes/Kohana/I18n.php
+++ b/system/classes/Kohana/I18n.php
@@ -117,7 +117,7 @@ class Kohana_I18n {
 				foreach ($files as $file)
 				{
 					// Merge the language strings into the sub table
-					$t = array_merge($t, Kohana::load($file));
+					$t = Arr::merge($t, Kohana::load($file));
 				}
 
 				// Append the sub table, preventing less specific language


errotan:
According to the documentation, this class uses one dimensional arrays to get translations in which case array_merge is enough.

errotan:
@saruman let's keep things simple, and use tools which do the job and no more. Switching from array_merge to Arr::merge increases the cyclomatic complexity of the load() method. Also introducing another hard dependency makes unit tests harder. Imagine that this change would be included and two years pass. Somebody looks at I18n documentation and this line in the code and says: "Hey, Arr::merge is not needed because translation arrays ['Hello, world!' => 'Bonjour, monde!'] are one dimensional. I make a PR to fix this." It gets accepted and boom, your application doesn't work. This change requires that you also change another parts of the I18n class and the documentation, so it would make senes to use multi dimensional arrays here.

saruman:
@errotan you right.

WinterSilence:
@saruman @errotan why not use http://php.net/manual/ru/function.array-merge-recursive ?

saruman:
@WinterSilence actually, this is a good idea and its simple.  What do you thing @errotan ?

errotan:
@saruman I already stated that `array_merge` is enough for merging one dimensional arrays.

piodro:
@saruman - Have you thought about extending the Kohana_I18n class?
You can overwrite the method "load()" with your own without changing the Core.

saruman:
@piodro yes, I already did that.  Thanks.

jstrobel:
If i read this right and the comments from @errotan are a good explanation, @saruman extendend the class for his own and there isn't a core-patch. So closing this.

PR #224, opened 2018-04-27
Commands out of sync

saruman:
Hello.
MySQL was throwing this error: Commands out of sync; you can't run this command now.

The MySQL client does not allow you to execute a new query where there are still rows to be fetched from an in-progress query. See https://dev.mysql.com/doc/refman/8.0/en/commands-out-of-sync.html.

I found the solution here:  https://stackoverflow.com/questions/14715889/strict-standards-mysqli-next-result-error-with-mysqli-multi-query

Now everything is working perfectly fine.  Please let me know.

### DIFF
--- a/modules/database/classes/Kohana/Database/MySQLi.php
+++ b/modules/database/classes/Kohana/Database/MySQLi.php
@@ -173,6 +173,9 @@ class Kohana_Database_MySQLi extends Database {
 				':query' => $sql
 			], $this->_connection->errno);
 		}
+		
+		do {}
+        	while (mysqli_more_results($this->_connection) AND mysqli_next_result($this->_connection));
 
 		if (isset($benchmark))
 		{


errotan:
Could you tell some use cases where this is required?

modules/database/classes/Kohana/Database/MySQLi.php
```diff
@@ -173,6 +173,9 @@ public function query($type, $sql, $as_object = FALSE, array $params = NULL)
 				':query' => $sql
 			], $this->_connection->errno);
 		}
+		
+		do {}
+        	while (mysqli_more_results($this->_connection) AND mysqli_next_result($this->_connection));
```
errotan: Using procedural style, should be like $this->_connection->more_results and $this->_connection->next_result

piotrbaczek: Please prepare DB/code where this is useful and an error is thrown/returned.

toitzi:
@saruman Can you please provide an example statement where this happends?

piotrbaczek:
This should be performed in your code, and not in the driver. Closing.

PR #225, opened 2018-04-28
refactor Cache_File to not cause side effects at construction

errotan:


errotan:
This does not alter the program behaviour, only does late initialization. The constructor should not do more than assigning variables. Merging this as there were no objections in the last 4 month.

PR #235, opened 2018-05-23
Added Cloudflare support when detecting Request::$client_ip

leonar15:
Also added test coverage for client_ip proxy types


errotan:
This header can be sent using a regular browser (eg. firefox). So this can be used to bypass some access control, like IP based banning?

PR #241, opened 2018-06-25
Update system/kohana/URL.php

klunker:
Add a check on subdomain into another logical tree, becouse subdomain don't added to url on subdomain pages.


WinterSilence:
@klunker mb better set subdomain once after this if/else?

errotan:
I rewrited this and committed into devel.

PR #242, opened 2018-07-05
Fix for #192 (original version by @Farazilu)

errotan:


errotan:
I'm doing this as a PR because this one needs discussion. What do you think?

Issue #19, opened 2017-01-12
Subdomains


piotrgolasz:
During last almost 3 years of using Kohana i got into some issues with subdomains.
I can without problems create a multi-application system, where I have different "application" folder for frontend, api, user, backend

by using .htaccess
`#API
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{HTTP_HOST} (www\.)?api\.*
RewriteRule .* api.php/$0 [PT,L]

#FRONTEND
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule .* index.php/$0 [PT,L]`

however there is no easy way to inject subdomains into URL::site(). This could be useful for example if being on my api.* subdomain I would like to refer to docs.*/issue/324 problem for the api.

neo22s:
I use at yclas.com subdomains without issues, I am not sure if I understood your issue. The problem is building URLs?

piotrgolasz:
yes, exactly

roman-huliak:
Kohana has very poor subdomains support. Applications using subdomains is many times too dificult. Subdomain can be dynamic or static  (dynamic=locations,categories,usernames). This need smart route filters, better route matcher.
On github is subdomain module for kohana but only for basic usage.

Issue #18, opened 2017-01-12
Userguide module not compatible with PHP7.1


neo22s:

Hyaena:
Where is this bug? Is it “[] operator not supported for strings” as seen in attached screenshot?

![userguide-error](https://cloud.githubusercontent.com/assets/383428/22156417/6d3992a8-df33-11e6-8375-e9ca11b8e3c9.png)

svenbw:
The error says that self::$_toc is a string and you can't push data on a string.
Initializing the variable as array should fix this... Or at least I think so

neo22s:
Thanks @Hyaena  https://github.com/koseven/koseven/pull/25

I need to test on PHP7.1 in deep still xD then I will close the issue.

Anyone wants to test too?

neo22s:
working!! https://docs.koseven.ga/

PHP7.1  thanks

PR #243, opened 2018-07-17
Feature/deprecate mcrypt

piotrbaczek:
* rewrite Encrypt engine to PHP 7.2 standard
* create Kohana_Engine_Sodium according to https://paragonie.com/book/pecl-libsodium/read/08-advanced.md#crypto-aead-aes256gcm
* clean mcrypt_ stuff from all Engines
* Mark Kohana_Engine_Mcrypt as deprecated

If there are errors, they are probably in code formatting.


errotan:
Maybe we should stick to php 7.0 compatibility for 3.3.x versions and consider php 7.1 or 7.2 for 3.4.x.

piotrgolasz:
TODO:
1) The return typing :?string is not compatible with 7.0 therefore has to be removed.
2) Unittests
3) Code formatting

piotrgolasz:
@errotan Do you maybe have time on hand to help finishing this?
https://github.com/koseven/koseven/pull/243

errotan:
@piotrgolasz sure. I'll look into this till the weekend.

piotrgolasz:
Some things left:
Engine mcrypt should be @deprecated, ie. in PHP 7.2 all MCRYPT_ constants are no longer available.
So i don't know if we shouldn't just mark Engine Mcrypt as deprecated, and just stop caring about tests aswell. Also the tests should use known encryption/decryption solutions as it was ;D

piotrgolasz:
Known AES KAT Vectors https://csrc.nist.gov/Projects/Cryptographic-Algorithm-Validation-Program/Block-Ciphers

piotrgolasz:
AES page was unavailable, there is a copy of KAT Vectors here: https://github.com/moserware/AES-Illustrated/blob/master/NIST%20Known%20Answer%20Test%20Vectors/CBCGFSbox256d.txt

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
```diff
@@ -0,0 +1,160 @@
...
+
+		if ($value === FALSE)
+		{
+			// Encryption failed
+			return NULL;
+		}
+
+		//Base64 encode binary data, otherwise they cannot be
+		//transformed into JSON.
+		$value = base64_encode($value);
```
piotrbaczek: This is unnecessary according to documentation https://github.com/paragonie/pecl-libsodium-doc/blob/master/chapters/08-advanced.md

just base64_encode($value) is more than enough.

neo22s:
I think this is almost done? and really great news BTW great JOB!!!!

toitzi:
@piotrbaczek  
Thanks for the implementation....if you don't mind i will take your code and add support for other encryption  algorithms supported by libsodium ( `ChaCha20 + Poly1305`, `ChaCha20 + Poly1305 (IETF Variant)`,` XChaCha20 + Poly1305`) actually already did the implementation i am gonna merge it into yours and post it here for discussion

piotrbaczek:
@toitzi What we need are actually Unittests with known KAT Vectors. If you could add those I would really appreciate as I simply don't have time to add those, hence the reason this PR is around so painstakingly long.

toitzi:
@piotrbaczek Added some KAT tests to your test file

toitzi:
I will add support and tests for other algorithms now

modules/encrypt/tests/EncryptMcryptTest.php
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * @dataProvider provider_encode_and_decode
+	 * @return void
+	 */
+	public function test_256_bit(string $encryptable)
+	{
+        $this->set_config([
+            'type' => Encrypt_Engine_Mcrypt::TYPE,
+            'cipher' => MCRYPT_RIJNDAEL_256,
```
piotrbaczek: MCRYPT_RIJNDAEL_256 is not AES256
AES256 is MCRYPT_RIJNDAEL_128 with 32bit key

toitzi: Yes you are right! thats one of the reasons why mcrypt is so bad:
https://paragonie.com/blog/2015/05/if-you-re-typing-word-mcrypt-into-your-code-you-re-doing-it-wrong#title.1.2

I will fix it with my next commit thanks!

piotrbaczek: Ok I fixed this

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
```diff
@@ -16,145 +16,145 @@
...
-		}
-
-		$decrypted = sodium_crypto_aead_aes256gcm_decrypt($value, '', $iv, $this->_key);
-
-		if ($decrypted === FALSE)
-		{
-			return NULL;
-		}
-
-		return $decrypted === FALSE ? NULL : $decrypted;
```
piotrbaczek: removed this double check that is unnecessary since L145

toitzi:
@piotrbaczek maybe we should add support for other algorithms before merging?

toitzi:
No i would simply add a configuration option like:

`'algorithm' =>Sodium_Algorithm_AES256`

in Engine_Sodium:

```
const Sodium_Algorithm_AES256 = 'aes256gcm';
const Sodium_Algorithm_ChaCha20 = 'chacha20poly1305';
const Sodium_Algorithm_ChaCha20IETF = 'chacha20poly1305_ietf';
const Sodium_Algorithm_XChaCha20 = 'xchacha20poly1305_ietf';
```
with according values, key length, iv size, encrypt and decrypt function can be built easily, if you want to i can update the code

toitzi:
@piotrbaczek no not like this, as i said you can easily get it from it's values, i will post an example of what i mean

> I would also like to stick to "S" in SOLID that a class should perform only single responsibility.

Yes sodium class will take 1 responsibility: decrypting and encrypting with libsodium (and all it's supported ciphers) We did same with mcrypt. I think putting them in seperate classes would make absolute no sense, but as i said i will show you an example of what i mean...i'll try to finish this till tomorrow

toitzi:
````
<?php
/**
 * The Encrypt Sodium engine provides two-way encryption of text and binary strings
 * using the [Sodium](http://php.net/sodium) extension, which consists of two
 * parts: the key and the cipher.
 *
 * The Key
 * :  A secret passphrase that is used for encoding and decoding
 *
 * @package    Kohana/Encrypt
 * @category   Security
 * @author     Koseven Team
 * @copyright  (c) Koseven Team
 * @license    https://koseven.ga/LICENSE.md
 */
class Kohana_Encrypt_Engine_Sodium extends Kohana_Encrypt_Engine
{
    use Traits_Encrypt_Iv;

    /**
     * Engine type
     * @var string
     */
    const TYPE = 'Sodium';

    /**
     * Supported Sodium Ciphers
     * @var string
     */
    const Cipher_AES256 = 'aes256gcm';
    const Cipher_ChaCha20 = 'chacha20poly1305';
    const Cipher_ChaCha20IETF = 'chacha20poly1305_ietf';
    const Cipher_XChaCha20 = 'xchacha20poly1305_ietf';

    /**
     * The size of the Initialization Vector (IV) in bytes
     * @var int
     */
    protected $_iv_size;

    /**
     * Sodium constructor.
     * @param  array  $config  Array with configuration
     * @throws Kohana_Exception
     */
    public function __construct($config)
    {
        parent::__construct($config);
        if ( ! extension_loaded('sodium'))
        {
            throw new Kohana_Exception('Sodium extension is not available');
        }

        if ( ! isset($config['cipher']) || $config['cipher'] === NULL)
        {
            // Add the default cipher
            $this->_cipher = Encrypt_Engine_Sodium::Cipher_AES256;
        }

        // Can you access AES-256-GCM? This is only available if you have supported hardware.
        if ($this->_cipher === Encrypt_Engine_Sodium::Cipher_AES256 && ! sodium_crypto_aead_aes256gcm_is_available()) {
            throw new Kohana_Exception('AES-256-GCM is not available on your hardware.');
        }

        $length = mb_strlen($this->_key, '8bit');

        $required_length = constant('SODIUM_CRYPTO_AEAD_'.strtoupper($this->_cipher).'_KEYBYTES');

        if ($length !== $required_length)
        {
            // No valid encryption key is provided!
            throw new Kohana_Exception('No valid encryption key is defined in the encryption configuration: length should be :required_length for :cipher, is: :current_length', [
                ':cipher'          => $this->_cipher,
                ':required_length' => $required_length,
                ':current_length'  => $length
            ]);
        }

        $this->_iv_size = constant('SODIUM_CRYPTO_AEAD_'.strtoupper($this->_cipher).'_NPUBBYTES');
    }

    /**
     * Encrypts
     * @param string $message Your message to be encrypted
     * @param string $iv
     * @return null|string
     */
    public function encrypt(string $message, string $iv)
    {
        $value = call_user_func(
            'sodium_crypto_aead_' . $this->_cipher . '_encrypt', $message, '', $iv, $this->_key
        );

        if ($value === FALSE)
        {
            // Encryption failed
            return NULL;
        }

        //Base64 encode binary data, otherwise they cannot be
        //transformed into JSON.
        $value = base64_encode($value);
        $iv = base64_encode($iv);

        $json = json_encode(compact('iv', 'value'));

        return !is_string($json) ? NULL : base64_encode($json);
    }

    /**
     * Decrypts the ciphertext
     * @param string $ciphertext Ciphertext to be decrypted
     * @return null|string
     */
    public function decrypt(string $ciphertext)
    {
        // Convert the data back to binary
        $decode = base64_decode($ciphertext);
        //Check if base64 decoding succeeded
        if ($decode === FALSE)
        {
            return NULL;
        }

        $data = json_decode($decode, TRUE);

        //check if json_decode succeeded
        if (is_null($data))
        {
            return NULL;
        }

        // If the payload is not valid JSON or does not have the proper keys set we will
        // assume it is invalid and bail out of the routine since we will not be able
        // to decrypt the given value. We'll also check the MAC for this encryption.
        if (!$this->valid_payload($data))
        {
            // Decryption failed
            return NULL;
        }

        $iv = base64_decode($data['iv']);
        if ($iv === FALSE)
        {
            // Invalid base64 data
            return NULL;
        }

        // Here we will decrypt the value. If we are able to successfully decrypt it
        // we will then unserialize it and return it out to the caller. If we are
        // unable to decrypt this value we will return NULL.
        $value = base64_decode($data['value']);

        if ($value === FALSE)
        {
            return NULL;
        }

        $decrypted = call_user_func(
            'sodium_crypto_aead_' . $this->_cipher . '_decrypt', $value, '', $iv, $this->_key
        );

        return $decrypted === FALSE ? NULL : $decrypted;
    }

    /**
     * Verify that the encryption payload is valid.
     * @param $payload
     * @return bool
     */
    protected function valid_payload($payload): bool
    {
        return is_array($payload) &&
            isset($payload['iv'], $payload['value']) &&
            strlen(base64_decode($payload['iv'], TRUE)) === $this->_iv_size;
    }
}
````

**Comment by piotrbaczek** [comment]
> Also short question (probably i am missing something) but:
> why did you put the create_iv in an extra class? Testing against KAT vectors should be possible even if it is not since you can pass an IV to de `encode` function

Do you mean this: https://github.com/koseven/koseven/pull/243/files#diff-83c1600fa14031150d59d125ec4d1c22

Because mcrypt uses mcrypt_create_iv() function and both openssl and sodium use random_bytes() function which is a common function.

**Comment by piotrbaczek** [comment]
@toitzi Can you try implementing your solution and writing tests to that?

**Comment by toitzi** [comment]
Yes i will implement it and write tests

**Comment by piotrbaczek** [comment]
//note to self:
Don't save the private key to private variable of the function, because it gets dumped, during t.ex. var_dump()
```
object(Kohana_Encrypt)#18 (2) { ["_engine":"Kohana_Encrypt":private]=> object(Encrypt_Engine_Sodium)#19 (4) { ["_iv_size":protected]=> int(12) ["_key":protected]=> string(32) "k"$.Q;FP+}^(k'br7e78s[e-7E)\F^SZ" ["_mode":protected]=> NULL ["_cipher":protected]=> NULL } ["_name"]=> string(7) "default" }
```

instead get the key when it's used. This is to prevent that accidental var_dump() won't dump the key on production

and also write more tests.

File: modules/encrypt/config/encrypt.php [hunk 4]
```diff
@@ -5,14 +5,19 @@
  * string   key     secret passphrase
  */
 
+use Kohana_Encrypt_Engine_Mcrypt as Mcrypt;
+use Kohana_Encrypt_Engine_Openssl as OpenSSL;
+use Kohana_Encrypt_Engine_Sodium as Sodium;
+use Kohana_Encrypt_Engine as Engine;
```
toitzi: Why do you do this? With correct naming this shold not be done. Please change

modules/encrypt/classes/Kohana/Encrypt/Engine.php
```diff
@@ -1,78 +1,91 @@
...
+    const CONFIG_TYPE = 'type';
 
-	/**
-	 * @var string Encryption key
-	 */
-	protected $_key;
+    /**
+     * @var string name of the key in configuration
+     */
+    const CONFIG_KEY = 'key';
```
toitzi: Don't assign default value to eky...many function check for that and will therefore not work.
e.G: 

Let's go through the flow (Lets pretend we are instancing `default` the first time.
1. You instance Encrypt Engine with `Encrypt::instance()`

Inside class `Encrypt`:

2. It assigns configuration name is `default`
3. Loads Configuration `default` from config file `encrypt`
4. Checks if COSNTANT  `Kohana_Encrypt_Engine::CONFIG_KEY` is set

Inside Kohana_Encrypt_Engine::CONFIG_KEY, we can see, that it has now a default value, therefore it will always return true. 

Same happens instancing `Kohana_Encrypt_Engine` inside `constructor` it checks if variable is empty...it isn't even if key is not set.

toitzi:
I did't look through all the changes, please be patient with me while new years eve 😄 
But thanks for the great work! WIsh you a happy new year!

modules/encrypt/classes/Kohana/Encrypt/Engine.php
piotrbaczek: No, no, no.
Those constants simply represent index in the $config array.
So before it was
```
return [
   'key' => 'SomeSecretKey'
];
$this->_key = $config['key'];
```

And now it is:
```
const CONFIG_KEY = 'key';
$this->_key = $config[self::CONFIG_KEY];
```

modules/encrypt/config/encrypt.php
piotrbaczek: Those are simply aliases for the class names. Because sometimes in the code it's easier to write Engine::CONFIG_KEY rather than Kohana_Encrypt_Engine::CONFIG_KEY, especially when you type it multiple times

modules/encrypt/classes/Kohana/Encrypt/Engine.php
toitzi: oh yes i see thanks!

modules/encrypt/config/encrypt.php
toitzi: Yes i know,but i don't get why doing it in Configuration file, did that no where else

toitzi:
Is there any benefit from using smth like `Engine::CONFIG_TYPE => OpenSSL::TYPE` in configuration instead of `'type'=> 'OpenSSL'` like we did before / in other modules (database, cache,..)

piotrbaczek:
> Is there any benefit from using smth like `Engine::CONFIG_TYPE => OpenSSL::TYPE` in configuration instead of `'type'=> 'OpenSSL'` like we did before / in other modules (database, cache,..)

If you use constant you are less likely to misspell or make an error since you use a constant. Also when you need to change something it's easier to find usages of a constant than searching for a text in code - in PHPStorm you click on a constant with CTRL and it shows you all the usages.

modules/encrypt/config/encrypt.php
piotrbaczek: Ok, that's not a problem. I can fix this.

toitzi:
@piotrbaczek 

> If you use constant you are less likely to misspell or make an error since you use a constant. Also when you need to change something it's easier to find usages of a constant than searching for a text in code - in PHPStorm you click on a constant with CTRL and it shows you all the usages.

Yes thats indeed true, but as we provide examples i think mistyping a configuration value is not that big  of an issue. I am also not that sure if we should generally use constants in configuration as there is kind of a plan to also allow other configuration types (yml, json, xml,...) see: #245

toitzi:
@piotrbaczek p.s: Something out of the topic:
I also use PHPStorm and i love it 😄 I recommend you the plugin:

https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-

it's awesome for static code analysis. If you want to i will send you my configuration 😄

piotrbaczek:
Do I need to change anything, or will you accept the review @toitzi ?

toitzi:
@piotrbaczek I think we should change the config, and not use constants. I think doing itlike everywhere else is more consistent. But that is just my opinion. NOTE: Your configuration will also break backwards compatibility, please keep in mind to write it down in 4.0.0 version info.

@neo22s @errotan @jstrobel What do you guys think?

P.s Bevor approving i will push the changes with support for all algorithms. I have time tomorrow to do this!

errotan:
I'm busy until february 15. Can only look into this after that.

neo22s:
I think better without constants, using constants as in the current config can break things if you do not have the module installed for example the constant does not exists

piotrbaczek:
> I think better without constants, using constants as in the current config can break things if you do not have the module installed for example the constant does not exists

That's a valid reason. I'll revert it tomorrow.

modules/encrypt/classes/Kohana/Encrypt.php
```diff
@@ -10,91 +9,93 @@
...
-        $this->_name = $name;
+    	// Get Driver Type
+		$config['type'] = $config['type'] ?? self::$default_engine;
 
         // Set the engine class name
-        $engine_name = 'Encrypt_Engine_' . ucfirst($config[Kohana_Encrypt_Engine::CONFIG_TYPE]);
+        $engine_name = 'Encrypt_Engine_' . ucfirst($config['type']);
 
         // Create the engine class
-        $this->setEngine(new $engine_name($config));
```
piotrbaczek: Because function setEngine only accepts Kohana_Encrypt_Engine, this was validation that passed class implements Kohana_Encrypt_Engine, which you removed, I don't know why. Can you revert this?

modules/encrypt/classes/Kohana/Encrypt.php
```diff
@@ -109,15 +110,16 @@ private function setEngine(Kohana_Encrypt_Engine $engine)
...
-     * @param string|null Initialization Vector $iv
+     * @param  string 		$message  Message to be encrypted
+     * @param  string|null 	$iv 	  Initialization Vector
      * @return null|string
+	 * @throws Exception
      */
     public function encode(string $message, string $iv = NULL)
     {
-        if (is_null($iv))
+        if ($iv === NULL)
```
piotrbaczek: This is actually the same thing but ok.

modules/encrypt/classes/Kohana/Encrypt/Engine/Mcrypt.php
```diff
@@ -20,90 +19,50 @@
...
      * @param array $config Array with configuration
      * @throws Kohana_Exception
      */
     public function __construct($config)
     {
-        if (!extension_loaded('mcrypt'))
+        if ( ! extension_loaded('mcrypt'))
         {
-            throw new Kohana_Exception('Sodium extension is not available');
+            throw new Kohana_Exception('Mcrypt extension is not available');
```
piotrbaczek: +1 Good eye

modules/encrypt/classes/Kohana/Encrypt/Engine/Mcrypt.php
```diff
@@ -176,42 +135,10 @@ public function create_iv(): string
...
-     *
-     * This method is a shim to make PHP 5.6 behave in a B/C way for
-     * legacy key padding when shorter-than-supported keys are used
-     *
-     * @param   string $key encryption key
-     * @param   string $cipher mcrypt cipher
-     * @param   string $mode mcrypt mode
-     * @return bool|string
-     */
-    protected function _normalize_key($key, $cipher, $mode)
```
piotrbaczek: Where did this function go? I don't want to modify anything in this engine, in case something breaks backwards compatibility. Let's just leave this as it is, ok?

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -10,83 +9,50 @@
...
- * @author     Kohana Team
- * @copyright  (c) Kohana Team
+ * @author     Koseven Team
+ * @copyright  (c) 2007-2012 Kohana Team
+ * @copyright  (c) 2016-2018 Koseven Team
  * @license    https://koseven.ga/LICENSE.md
  */
 class Kohana_Encrypt_Engine_Openssl extends Kohana_Encrypt_Engine
 {
-    use Traits_Encrypt_Iv;
```
piotrbaczek: Code looks fine, this method can be overwritten in mcrypt, that's not a problem. I just wanna make sure you're not one of those people that never use traits, because someone told you they're bad (I have this kind of people at my work) 😄

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -10,83 +9,50 @@
...
-        if (empty($config[self::CONFIG_CIPHER]))
-        {
-            // Add the default cipher
-            $this->_cipher = static::AES_256_CBC;
-        }
-        else
-        {
-            $this->_cipher = $config[self::CONFIG_CIPHER];
-        }
+		$this->_cipher = $config['cipher'] ?? self::AES_256_CBC;
```
piotrbaczek: Hmm actually wondering if this is the same thing as $config['mode'] in mcrypt?

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -10,83 +9,50 @@
...
-                {
-                    // No valid encryption key is provided!
-                    throw new Kohana_Exception('No valid encryption key is defined in the encryption configuration: length should be 32 for AES-256-CBC');
-                }
-                break;
-            default:
-                // No valid encryption cipher is provided!
-                throw new Kohana_Exception('No valid and safe encryption cipher is defined in the encryption configuration.');
-        }
+		$this->valid_key_length($required_length);
```
piotrbaczek: left space

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
```diff
@@ -138,36 +98,18 @@ public function decrypt(string $data)
...
         // assume it is invalid and bail out of the routine since we will not be able
         // to decrypt the given value. We'll also check the MAC for this encryption.
-        if (!$this->valid_payload($data))
-        {
-            // Decryption failed
-            return NULL;
-        }
-
-        if (!$this->valid_mac($data))
+        if (!$this->valid_payload($data) || !$this->valid_mac($data) || !$iv = base64_decode($data['iv']))
```
piotrbaczek: Please let's not use same line assign and comparison. Assign $iv one line before if statement, thank you.

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
```diff
@@ -35,26 +47,31 @@ class Kohana_Encrypt_Engine_Sodium extends Kohana_Encrypt_Engine
...
     public function __construct($config)
     {
-        if (!extension_loaded('sodium') || !sodium_crypto_aead_aes256gcm_is_available())
+        if ( ! extension_loaded('sodium'))
         {
             throw new Kohana_Exception('Sodium extension is not available');
         }
 
+        // Check if cipher is set, otherwise fallback to AES 256 + GCM
+		if ( ! isset($config['cipher']) || $config['cipher'] === NULL)
```
piotrbaczek: $config['cipher'] is this same thing as $config['mode'] in mcrypt?

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
```diff
@@ -64,87 +81,53 @@ public function __construct($config)
...
-        }
-
-        $iv = base64_decode($data['iv']);
-        if ($iv === FALSE)
-        {
-            // Invalid base64 data
+		// If the payload is not valid JSON or does not have the proper keys set we will
+		// assume it is invalid and bail out of the routine since we will not be able
+		// to decrypt the given value. We'll also check the MAC for this encryption.
+        if (
```
piotrbaczek: formatting

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
```diff
@@ -64,87 +81,53 @@ public function __construct($config)
...
+			$decode === FALSE || ($data = json_decode($decode, TRUE)) === NULL ||
```
piotrbaczek: Can we not assign and compare in the same line again. please?

piotrbaczek: Actually this if statement is very hard to debug, can we go back to previous if statements?

modules/encrypt/config/encrypt.php
```diff
@@ -2,20 +2,20 @@
...
-        Kohana_Encrypt_Engine::CONFIG_KEY => NULL,
-        //Additional OpenSSL configuration
-        Kohana_Encrypt_Engine::CONFIG_CIPHER => Kohana_Encrypt_Engine_Openssl::AES_256_CBC,
-    ],
-//    'sodium' => [
-//        Kohana_Encrypt_Engine::CONFIG_TYPE => Kohana_Encrypt_Engine_Sodium::TYPE,
-//        Kohana_Encrypt_Engine::CONFIG_KEY => NULL,
-//    ],
+	'default' => [
+		'type'    => 'sodium',
```
piotrbaczek: Are we sure sodium is the new default and not openssl?

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
piotrbaczek: Also formatting

modules/encrypt/classes/Kohana/Encrypt/Engine/Mcrypt.php
toitzi: I can but this function was only needed for php <= 5.6. Since Koseven is only PHP 7.0 Compatible this is not neccessary, but if you want i can revert

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
toitzi: Hahaha no basicly i like them and also use them a lot. But since this is a common function it can be added to the parent class and overwritten if needed.

modules/encrypt/config/encrypt.php
toitzi: Actually i don't care about that i just tested and my commit wasn't "final" but yeah let's switch it to openssl :)

modules/encrypt/classes/Kohana/Encrypt/Engine/Sodium.php
toitzi: Yes

modules/encrypt/classes/Kohana/Encrypt.php
toitzi: You didn't validate it you just required it in the function parameters. This would throw a php error if engine is not an instance - which is not good. So basicly your function was just a setter which didn't perform any actions. According to [google](https://web.archive.org/web/20140625191431/https://developers.google.com/speed/articles/optimizing-php) this reduces speed and is considered a bad practice

But i agree we need to validate... i will add a check if the class is availiable/exists AND after calling if is an instance of Kohana_Encrypt_Engine. If not it will throw a Kohana Exception stating that the engine is not valid.

modules/encrypt/classes/Kohana/Encrypt/Engine/Mcrypt.php
piotrbaczek: Ok in that case this can be removed.

modules/encrypt/classes/Kohana/Encrypt/Engine/Openssl.php
toitzi: No. according to documentation and code here is the difference:
'cipher' determines how the encryption is mathematically calculated e.g: AES-128
'mode' etermines how the encrypted data is written in binary form e.g: nofb

toitzi:
@piotrbaczek I am now done with unittests. Everything seems to work fine.
Please do a final review. Then i will **finally** 😄 merge this PR into branch 4.0.0

toitzi:
@neo22s According to our Contributing guide, branch 4.0 only is created / shall be used when 4.0 releases and develop is not for production use (of course) and is branch for new features, etc... i will merge it there...is this ok?

modules/encrypt/classes/Kohana/Encrypt.php
```diff
@@ -75,7 +75,7 @@ public static function instance(string $name = NULL, array $config = NULL) : sel
                 $config = Kohana::$config->load('encrypt')->$name;
             }
 
-            if (!isset($config['key']))
+            if (!isset($config['key']) || $config['key'] === NULL || empty($config['key']))
```
piotrbaczek: empty is enough, returns true if null

modules/encrypt/classes/Kohana/Encrypt/Engine.php
```diff
@@ -96,6 +92,7 @@ protected function valid_key_length($expected) {
 
     /**
      * Override __debugInfo function to not display key in var_dump
+	 * @codeCoverageIgnore
```
piotrbaczek: formatting

modules/encrypt/tests/EncryptSodiumTest.php
```diff
@@ -77,6 +77,83 @@ public function test_decrypt_invalid()
 		$this->assertNull($encrypt->decode(base64_encode('asdasd')));
 	}
 
+	/**
+	 * Test with invalid Key Length and expect exception
+	 * @expectedException Kohana_Exception
+	 * @throws Kohana_Exception
+	 */
+	public function test_invalid_key_length() {
```
piotrbaczek: Those 3 functions beneath perform no asserts

toitzi: no if you look closley you can see in the doc comments expectedException... that's the assert they perform. As we test invalid configuration they expect an exception. Tested it with php unit and test passes. If expectedExceptions wouldn't count as assertion phpunit would throw an error: "No assertion performed" which it doesn't . Therefore this is valid (also according to official PHPUnit Documentation)

modules/encrypt/classes/Kohana/Encrypt/Engine.php
toitzi: fixed

modules/encrypt/classes/Kohana/Encrypt.php
toitzi: fixed

PR #244, opened 2018-07-26
Behavior updates

svenbw:
Fixed some indents.
Added option to switch off the verification of the GUID(/UUID), when inserting a lot of columns this check decreased the insert performance.
Uniqueness should be set on the database column.


modules/orm/classes/ORM/Behavior/Guid.php
```diff
@@ -1,113 +1,148 @@
...
+		}
+
+		// Try to create a new GUID
+		$query = DB::select()->from($model->table_name())
+			->where($this->_guid_column, '=', ':guid')
+			->limit(1);
+
+		while (empty($current_guid))
+		{
+			$current_guid = sprintf('%04x%04x-%04x-%04x-%04x-%04x%04x%04x',
```
piotrgolasz: This should be an independent function - in a complete new class, so it can be reused.

svenbw: OK, there are UUID modules available, we can include this module by default.
Or I add a class to the Kohana core, since it's very simple I think this is the best option.

svenbw: @piotrgolasz I wrote a class for the core, but it might cause some issues on existing installations. "UUID" as a class is a commonly used name, putting it in the core is a risk.
Or I create an `ORMUUID` class, which is probably unique. What do you think?

piotrbaczek: ORM UUID type would be best imho.

toitzi:
@svenbw @piotrbaczek What's the current state / what needs to be done here?

toitzi:
Fixed with #319

PR #249, opened 2018-08-07
Fixed error handlers (strict compliance with PHP 7)

### LINKED ISSUE #248: Error handlers in strict accordance with PHP 7
Koseven installer requires PHP version at least 7.0:
https://github.com/koseven/koseven/blob/devel/public/install.php#L136

Tests for PHP 5 are not performed:
https://github.com/koseven/koseven/blob/devel/.travis.yml#L4

We can replace the error handlers in strict accordance with PHP 7?

http://php.net/manual/en/function.set-exception-handler.php

What do you think of it?

bkosun:
Fixed #248

### DIFF
--- a/modules/minion/classes/Kohana/Minion/Exception.php
+++ b/modules/minion/classes/Kohana/Minion/Exception.php
@@ -16,26 +16,26 @@ class Kohana_Minion_Exception extends Kohana_Exception {
 	 * Should this display a stack trace? It's useful.
 	 *
 	 * @uses    Kohana_Exception::text
-	 * @param   Exception   $e
+	 * @param   Throwable  $t
 	 * @return  boolean
 	 */
-	public static function handler($e)
+	public static function handler(Throwable $t)
 	{
 		try
 		{
 			// Log the exception
-			Kohana_Exception::log($e);
+			Kohana_Exception::log($t);
 
-			if ($e instanceof Minion_Exception)
+			if ($t instanceof Minion_Exception)
 			{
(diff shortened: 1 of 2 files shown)


neo22s:
@errotan we can add this for 4.0.0 and then update the docs for migration.

Should I merge?

errotan:
Before merging can we rename the master branch? I written a roadmap a while ago and members agreed on it on telegram: https://github.com/koseven/koseven/blob/devel/ROADMAP.md It is ok if you think it is not required, but than roadmap needs to be edited.

neo22s:
Sure!  I think gets more clear ;)

Rename current to 3.3 and we create a new one based on devel called master?

thanks

errotan:
Nope. Master is not needed as the name does not designate the purpose of the branch.

neo22s:
@errotan sorry I meant 4.0 instead of master I do not know why I wrote master :S

"Rename current to 3.3 and we create a new one based on devel called 4.0" I wanted to say hahah

toitzi:
@neo22s Is there anything todo here?

neo22s:
We need to update the docs, explaining this change, on migrate from 3.x to 4.x

toitzi:
Before we release 4.0 we can write a guide or something if needed. To keep track and don't forget it i added it to 4.0's [Changelog](https://github.com/koseven/koseven/blob/v4.0/CHANGELOG.md).
Let's collect all changes there and write examples / docs if needed when they are final

PR #256, opened 2018-09-13
Implemented random_bytes for security token generation

partnerizemark:


partnerizemark:
@errotan updated with requested syntax - i also noticed that if existing token was present and $new was True that if the random_bytes function was missing the previous token would be returned (v. bad) changed this to setting $token = NULL then using a finally block to ensure the token is generated on NULL

system/classes/Kohana/Security.php
```diff
@@ -47,17 +47,36 @@ public static function token($new = FALSE)
 
 		if ($new === TRUE OR ! $token)
 		{
-			// Generate a new unique token
-			if (function_exists('openssl_random_pseudo_bytes'))
+			$token = NULL;
+
+			try
+			{
+				if (function_exists('random_bytes')) {
```
bkosun: > Curly brackets are placed on their own line, indented to the same level as the control statement.

https://docs.koseven.ga/guide/kohana/conventions#curly-brackets

system/classes/Kohana/Security.php
```diff
@@ -47,17 +47,36 @@ public static function token($new = FALSE)
...
+				// Random bytes function is available but no sources of randomness are available
+				// so rather than allowing the exception to be thrown - fall back to other methods.
+				// @see http://php.net/manual/en/function.random-bytes.php
 			}
-			else
+			finally
 			{
-				// Otherwise, fall back to a hashed uniqid
-				$token = sha1(uniqid(NULL, TRUE));
+				if (! $token)
```
bkosun: > There should be one space after statement name, followed by a parenthesis. The ! (bang) character must have a space on either side to ensure maximum readability. Except in the case of a bang or type casting, there should be no whitespace after an opening parenthesis or before a closing parenthesis.
https://docs.koseven.ga/guide/kohana/conventions#parentheses

system/classes/Kohana/Security.php
```diff
@@ -83,25 +102,25 @@ public static function check($token)
 	{
 		return Security::slow_equals(Security::token(), $token);
 	}
-	
-	
-	
+
+
+
```
bkosun: Unnecessary line breaks

partnerizemark:
@bkosun see latest commit for implementations of the conventions and coding styles advised.

errotan:
Changed it a little bit: 6d42bb5

partnerizemark:
@errotan example present in forked repo

partnerizemark:
@errotan apologies ignore my comment, i now see you replaced the $token = with returns.

PR #259, opened 2018-10-03
Update View, Core, UTF8 classes

WinterSilence:


system/classes/Kohana/Core.php
```diff
@@ -233,7 +233,7 @@ public static function init(array $settings = NULL)
 					// Set permissions (must be manually set to fix umask issues)
 					chmod($settings['cache_dir'], 0755);
 				}
-				catch (Exception $e)
+				catch (Throwable $e)
```
bkosun: Why do you use Throwable instead of Exception?

system/classes/Kohana/Core.php
```diff
@@ -717,13 +717,10 @@ public static function list_files($directory = NULL, array $paths = NULL)
 							}
 						}
 					}
-					else
+					elseif ( ! isset($found[$key]))
```
bkosun: What you do not like the previous solution?

system/classes/Kohana/Core.php
```diff
@@ -942,7 +939,7 @@ public static function error_handler($code, $error, $file = NULL, $line = NULL)
 		{
 			// This error is not suppressed by current error reporting settings
 			// Convert the error into an ErrorException
-			throw new ErrorException($error, $code, 0, $file, $line);
+			throw new ErrorException($error, $code, $code, $file, $line);
```
bkosun: Why should the $code and $severity parameters be the same?
http://php.net/manual/en/errorexception.construct.php

system/classes/Kohana/Core.php
```diff
@@ -983,7 +980,7 @@ public static function shutdown_handler()
 			ob_get_level() AND ob_clean();
 
 			// Fake an exception for nice debugging
-			Kohana_Exception::handler(new ErrorException($error['message'], $error['type'], 0, $error['file'], $error['line']));
+			Kohana_Exception::handler(new ErrorException($error['message'], $error['type'], $error['type'], $error['file'], $error['line']));
```
bkosun: Why should the $code and $severity parameters be the same?
http://php.net/manual/en/errorexception.construct.php

system/classes/Kohana/UTF8.php
```diff
@@ -138,11 +147,13 @@ public static function strip_non_ascii($str)
...
-	 * @param   string  $str    string to transliterate
-	 * @param   integer $case   -1 lowercase only, +1 uppercase only, 0 both cases
+	 * @param   string  $str           string to transliterate
+	 * @param   integer $case          -1 lowercase only, +1 uppercase only, 0 both cases
+	 * @param   array   $lower_accents lower accents
+	 * @param   array   $upper_accents upper accents
 	 * @return  string
 	 */
-	public static function transliterate_to_ascii($str, $case = 0)
+	public static function transliterate_to_ascii($str, $case = 0, array $lower_accents = NULL, array $upper_accents = NULL): string
```
bkosun: Why do you explicitly specify the return type? As far as I understand, in Koseven this can be done only in special cases.

system/classes/Kohana/View.php
```diff
@@ -233,19 +228,19 @@ public function __toString()
 	{
 		try
 		{
-			return $this->render();
+			return (string) $this->render();
```
bkosun: The render() method returns a string.

system/classes/Kohana/View.php
```diff
@@ -255,12 +250,13 @@ public function __toString()
...
 	 * @param   string  $file   view filename
-	 * @return  View
+	 * @return  self
 	 * @throws  View_Exception
 	 */
 	public function set_filename($file)
 	{
-		if (($path = Kohana::find_file('views', $file)) === FALSE)
+		$path = file_exists($file) ? $file : Kohana::find_file('views', $file);
+		if (!$path)
```
bkosun: https://docs.koseven.ga/guide/kohana/conventions#parentheses

system/classes/Kohana/Core.php
WinterSilence: Need catch Error (Error/Exception implements Throwable).

system/classes/Kohana/Core.php
WinterSilence: its shortcut

system/classes/Kohana/Core.php
WinterSilence: > While the severity can be any integer value, it is intended that the error constants be used. 

http://us2.php.net/manual/en/errorexception.construct.php

system/classes/Kohana/View.php
WinterSilence: render method at errors can be return void/null

system/classes/Kohana/UTF8.php
WinterSilence: we cant update typing for all classes, can do it step by step.

bkosun: To implement such changes, you need to add a new feature branch and work there, and then merge the branches.

system/classes/Kohana/View.php
bkosun: https://github.com/koseven/koseven/blob/ec1778f1fe9175fdb07f1babfedcd37716cc4c67/system/classes/Kohana/View.php#L359
https://github.com/koseven/koseven/blob/ec1778f1fe9175fdb07f1babfedcd37716cc4c67/system/classes/Kohana/View.php#L74

According to the documentation
```
/**
 * Get current buffer contents and delete current output buffer
 * @link http://php.net/manual/en/function.ob-get-clean.php
 * @return string the contents of the output buffer and end output buffering.
 * If output buffering isn't active then false is returned.
 * @since 4.3.0
 * @since 5.0
 */
function ob_get_clean () {}
```

system/classes/Kohana/Core.php
bkosun: ?

WinterSilence: @bkosun see  example by php team: severity = code, code = 0, i dont set code = 0 because its can break users code.

system/classes/Kohana/View.php
WinterSilence: @bkosun search problems then method return not string harder than fix returned value.

system/classes/Kohana/Core.php
bkosun: Unfortunately I did not find this note on the page indicated.

WinterSilence: >  severity The severity level of the exception.
Note: While the severity can be any integer value, it is intended that the error constants be used.

WinterSilence: @bkosun http://us2.php.net/manual/en/class.errorexception.php

bkosun: But why do you think `$code` should be equal to `$severity?`

You yourself quoted:
> Note: While the severity can be any integer value, it is intended that the error constants be used.

These constants are:
http://php.net/manual/en/errorfunc.constants.php#errorfunc.constants

The `$code` parameter can contain any value so that you can correctly handle the error.  

For example, if you have a child `Exception` class that handles a MySQL database error, then `$code` may be a MySQL error code.

WinterSilence: @bkosun  >  Is not it? 

nope., because severity added in ErrorException.

PR #260, opened 2018-10-07
remove unused variables

errotan:
found using phpcs-variable-analysis


modules/auth/classes/Kohana/Auth/File.php
```diff
@@ -34,6 +34,11 @@ public function __construct($config = [])
 	 */
 	protected function _login($username, $password, $remember)
 	{
+		if ($remember)
```
piotrgolasz: +1

PR #263, opened 2018-10-22
Pass directory param in Pagination constructor

piotrbaczek:


neo22s:
only issue I see that the Directory may not be added in the pagination. But does not hurt anyone. good for me.

PR #264, opened 2018-10-24
Add JSON field type to MySQLi Driver

uBizzy:
"JSON" is an alias field type for "LONGTEXT". This edit make MySQLi Driver compatible to use "JSON" field type.

### DIFF
--- a/modules/database/classes/Kohana/Database/MySQLi.php
+++ b/modules/database/classes/Kohana/Database/MySQLi.php
@@ -222,6 +222,7 @@ class Kohana_Database_MySQLi extends Database {
 			'integer unsigned'          => ['type' => 'int', 'min' => '0', 'max' => '4294967295'],
 			'longblob'                  => ['type' => 'string', 'binary' => TRUE, 'character_maximum_length' => '4294967295'],
 			'longtext'                  => ['type' => 'string', 'character_maximum_length' => '4294967295'],
+			'json'                      => ['type' => 'string', 'character_maximum_length' => '4294967295'],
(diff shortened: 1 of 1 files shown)


piotrbaczek:
@uBizzy can you also add JSON to ORM, so that the field that is JSON will get json_encoded before saving and json_decoded after pull?

uBizzy:
@piotrbaczek , do you think it is really necessary? I see no obligation to have it that way because each one can decide which to use. The only thing I envision here is:

1. Return only a json on SELECT, then the user decide if he want to decode or not, or pass it as an option
2. On INSERT / UPDATE verify if it is an array and apply the json_encode. Any way, if it is not a valid json the SQL will return an error with a non valid value for this field and it then Kohana catch that error.

What are your thoughts?

bkosun:
@piotrbaczek,

You can declare properties to be serialized, here’s a simple example:

```
class Model_User extends ORM
{
    protected $_serialize_columns = [
        'column_name_1',
        'column_name_2',
        'column_name_3',
    ];
}
```

The json_encode / json_decode functions will be automatically applied to the specified properties.

https://github.com/koseven/koseven/blob/6c1c48eba774104681e1ba0bebcd06e8e315140a/modules/orm/classes/Kohana/ORM.php#L1923
https://github.com/koseven/koseven/blob/6c1c48eba774104681e1ba0bebcd06e8e315140a/modules/orm/classes/Kohana/ORM.php#L1932

uBizzy:
Any feedback about this pull request?

Issue #282, opened 2018-11-20
Proposal to use Koseven name in classes


neo22s:
For example, instead of class Kohana_Core => Koseven_Core.

Also replace Kohana folders for Koseven.

We could keep the Kohana name so migration it's easy and backwards compatible, although doing a find replace it's not complex.

I know does not add any functionality but I think by releasing 4.0.0 maybe it's worth to use the new branding.

Of course always we will keep credit to all the kohana team work.

What's your opinion?

jstrobel:
Think that too. With release 4 we can rename it.

piotrbaczek:
I'm against it. It's a lot of very tedious work and we get nothing in return. I would rather spend that time developing or improving software, writing unittest event to get us from pity 35% to a 100%.

neo22s:
@piotrbaczek of course this has low priority, but at some point will need to be done, it's super strange and confusing for anyone that wants to start a new project using Koseven, that needs to type Kohana :S

If we release 4.0.0 it's because we also want new people to start using the project I think no?

neo22s:
We can just simply extend the kohana class to have another one called koseven... and ready, but having all folders called kohana it's ugly as hell.

We are plenty of people, if we organize our effort properly we can make it.

I know I have not been there for almost a year but seems that now after January I will be more free (december holidays xD).

I am not the best developer ever I can handle myself, but I have been successfully leading teams and small companies with almost 0 budget for years, also an open source project which is what I make a living from.

If we all go to the same path together we can achive anything we want.

I believe that Koseven it's a really powerful framework as valid as any other one out there and we can improve that.

Also I am aware of some tensions, probably the way I write. I like to be direct and honest and I love to listen to everyone. If you knew me in person you may put a tone to my writing. It's the bad part of remote working, you do not get to know the people fully.

Thanks to all of you again.

uBizzy:
I agree with @piotrbaczek but I fully understand @neo22s 's vision. If we ever release a Koseven v4 for sure this would be a big improvement but not the only one.

toitzi:
This may require time...but i got some "spare time ;)". If somone just could explain to me how you want this to be done (maybe with a small example)..I can do it

svenbw:
@toitzi That's the idea; it would be even better if this could be "scripted", if a local repo of user uses also the Kohana classes the code can be updated automatically.
I wrote some code which renames the directories, for classes I assume that some kind of reflection / token parsing is required to cover everything.

[...]

PR #285, opened 2018-11-22
Renamed Kohana Classes to Koseven and created a compatibility module

toitzi:
See Issue #282


neo22s:
Should we merge this to Devel and test?

toitzi:
Don't merge...issues occur when you try to include Kohana classes directly.

eg.: `require SYSPATH.'classes/kohana.php';`

We should  not use it as model for full compatiblity we should keep it in "/system"

Further testing over the weekend.

toitzi:
@neo22s 
Think i found a solution:

add the following to `Koseven_Core`:

````
/**
 * @var boolean  Kohana Compatibility mode
 */
public static $compatibility = TRUE;
````

after `spl_autoload_register` inside `bootstrap.php`

````
if (Koseven::$compatibility AND is_file(APPPATH.'classes/Kohana'.EXT))
{
    require APPPATH.'classes/Kohana'.EXT;
}
````
We can now also autoload the module if `Koseven::$compatibility` is true.

And last but not least add the following section inside `Koseven::find_file` function:
````
...
/**
 *     NOTE: This check is only needed if we are PRE-Initialization and in Compatibility mode
 *     Only performing before Initialization makes sure that no `stpos` operations, etc get called
 *     on custom classes and without beeing necessary
 **/ 
if (!Koseven::$_init AND Koseven::$compatibility AND strpos($path, 'kohana') !== FALSE) 
{                 
     $found = MODPATH.'kohana'.DIRECTORY_SEPARATOR.$path;
     if (!is_file($found))
     {
          $found = FALSE;
     }
}

....
````
This makes sure we can keep it as a module, cause if kohana classes are loaded
PRE-Initializaton (like the one from @neo22s  in his project yclass) - which means even before modules are loaded - this function makes sure to check them inside the kohana module folder

neo22s:
What do you all think guys? 

I mean for me when I do the migration I have no issue into renaming all the project. Will take maybe few minutes but that's all.

PR #289, opened 2018-11-23
Koseven Rename with Compatibility Module

toitzi:
See #285 and #282

### DIFF
--- a/application/bootstrap.php
+++ b/application/bootstrap.php
@@ -2,7 +2,7 @@
 
 // -- Environment setup --------------------------------------------------------
 
-// Load the core Kohana class
+// Load the core Koseven class
 require SYSPATH.'classes/Koseven/Core'.EXT;
 
 if (is_file(APPPATH.'classes/Koseven'.EXT))
@@ -33,7 +33,7 @@ date_default_timezone_set('America/Chicago');
 setlocale(LC_ALL, 'en_US.utf-8');
 
 /**
- * Enable the Kohana auto-loader.
+ * Enable the Koseven auto-loader.
  *
  * @link http://kohanaframework.org/guide/using.autoloading
  * @link http://www.php.net/manual/function.spl-autoload-register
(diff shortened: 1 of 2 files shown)


neo22s:
@toitzi I will take your devel branch test monday, thanks a loooot! :D

toitzi:
@neo22s 

I think i know the solution pretty sure it's number 3 but also check 1&2 pls...

1. Add this under your `spl_autload` in `bootstrap.php` function....should be there if you merged correctly, but please check it again.
````
if (Koseven::$compatibility AND is_file(APPPATH.'classes/Kohana'.EXT))
{
    // Application extends the core
    require APPPATH.'classes/Koseven'.EXT;
}
````

2. Change all the `requires` on top of your `bootstrap.php` (Should also be already done if you merged correctly....but check it again..Has to look like:
````
require SYSPATH.'classes/Koseven/Core'.EXT;
if (is_file(APPPATH.'classes/Koseven'.EXT))
{
	// Application extends the core
	require APPPATH.'classes/Koseven'.EXT;
}
else
{
	// Load empty core extension
	require SYSPATH.'classes/Koseven'.EXT;
}
````
3. You renamed the constant `MODPATH` in your project to `KOMODPATH` the compatibility function looks for it inside `MODPATH`...you either move the module there OR you change `MODPATH` in
`system/classes/Koseven/Core.php` line 635 to `KOMODPATH`

toitzi:
Tested it with "yclas" from your github: https://github.com/yclas/yclas/
worked like charm...if you still get an error,
please provide me the exact error message and your `bootstrap.php`

neo22s:
@toitzi I am testing on another project, I will try later again, finishing the working day ;) pretty sure it is my fault.

neo22s:
What we need it's like some short of super small step by step:

1- activate module
2- in bootstrap.php do bla bla bla

Not only for me but for all of us and new users heheh

toitzi:
@neo22s yes i think that's a great idea...something like a small "upgrade" guide...doesn't need much information, just a short step by step

WinterSilence:
I think easy solution its: replace `Kohana` to `Koseven` in all php files of your project :)

Sorry, but easy replace `Kohana` to `Koseven` in not work. Also need replace code via :
~~~
// get class name without 'Kohana_'
$class = substr(get_class($this), 7);
~~~

toitzi:
@WinterSilence 

Can you give detailed information, why it didn't work?

WinterSilence:
@toitzi i talk about external/users code

toitzi:
@WinterSilence Can you provide an example code where this could possibly break the application of someone?

toitzi:
My approach now would be:
Close this PR,
Create a new one, with doing almost same but using KO7 instead of koseven and do it a bit more careful. I couldn't really understand what @WinterSilence means by it will break external code. As all classes are still available. the only reason i see this will break code if someone tries to include framework files directly. But i don't think that's a big issue / something many people have / something that isn't hard to change, but i am not sure....what do you say @neo22s @jstrobel 

Also i don't like the idea of simply string manipulate inside autoloader for following reasons:
1. No Code completion in IDE or anything like that.
2. Since we want to deprecate "Kohana" functions / callings and remove it later, this wouldn't do the thing

neo22s:
Sounds great by me! Thanks

jstrobel:
@toitzi Sound good for me too. Thanks a lot.

Issue #252, opened 2018-08-21
Redis cache


neo22s:
Same as memcached. But using redis.

We can use as base https://github.com/bkhrupa/kohana-rediscache

neo22s:
I do not think there's need of making many changes. I have adapted already the code, needs to be tested in production (tests working) and staging working fine. as soon as I am 100% sure works good I will do a PR. Basically it's copy pasting the module making few changes so it's is compatible.

chrisgo:
I have this one too (super old with no tests unfortunately) with tagging support https://github.com/chrisgo/kohana-cache-redis

PR #291, opened 2018-11-30
Redis

toitzi:
#252


modules/cache/tests/cache/RedisTest.php
```diff
@@ -0,0 +1,46 @@
+<?php
+include_once(Kohana::find_file('tests/cache', 'CacheBasicMethodsTest'));
```
piotrbaczek: This include is sort of weird. I see that other cache test classes use it aswell, but i think with proper naming we should be able to use Kohana filesystem, right?

modules/cache/classes/Kohana/Cache/Redis.php
```diff
@@ -0,0 +1,282 @@
...
+    public function __construct(array $config)
+    {
+        if ( ! extension_loaded('redis'))
+        {
+            throw new Cache_Exception(__METHOD__.' Redis PHP extension not loaded!');
+        }
+
+        parent::__construct($config);
+
+        $this->_redis = new Redis();
```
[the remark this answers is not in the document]
toitzi: no..don't think so

modules/cache/tests/cache/RedisTest.php
toitzi: Guess you are right...i just kept it there because as you said it is in other cache calsses used aswell

toitzi: Oh i see,
no this includeis neccessary, because `CacheBasicMethodsTest`doesn't inside `tests` folder doesn't follow kohana naming convetion

toitzi:
Added all tests, neccessary, my IDE (PHPStorm) also shows me that coverage is fine.
Can someone explain to me why "coveralls" still says coverage in my Redis file is 0.00% ?
Am i missing something?

toitzi:
Will rewrite tests, and "fix" code a bit, will then ask for review to merge

toitzi:
@piotrbaczek @errotan done, can this be merged?

toitzi:
@neo22s can i merge this?

neo22s:
@toitzi I have not reviewed :( I am not in home until 1st january. Looks good to me though.

PR #294, opened 2018-12-14
Fix bug in ORM::values()

WinterSilence:
Remove PK value only if model loaded (PK not always auto increment).
~~~
// model not loaded
$model = ORM::factory('Foo');
// but id removed because it's PK
$model->values(['id' => 123, 'name' =>'Anton']);
// id column not 'autoincrement', as result - error
$model->save();
~~~


saruman:
You saved me @WinterSilence, thank you.

PR #298, opened 2018-12-17
Travis CI Configuration Update

toitzi:
Updated our travis configuration:

Updated distribution from Ubuntu 14.04 to 16.04
Added the following services (which run with their default configuration) and can be used for unittesting:

```
redis-server
memcached
mysql
```
Created a database `unittest`.

Added the following PHP Extensions (required to run some unittests - which will also incrase our coverage if we change a few lines):

```
redis
imagick
apcu
memcached
```

Last but not least added:

`composer self-update` (to keep composer itself up to date)
`composer validate` (which will check if composer.json is valid)

Devel branch passes build test with this configuration, so i am pretty sure this can be merged into devel!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,8 @@
 # Project language is PHP7
 language: php
 
-# Test on multiple Operating systems
-os:
-  - linux
-
-# Linux OS should be trusty (16.04)
-matrix:
-  include:
-    - os: linux
-      dist: xenial
+# Test on Ubuntu Xenial (16.04)
+dist: xenial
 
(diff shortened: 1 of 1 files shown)


toitzi:
@errotan yes it does, can you just give me a quick explanation why you think this is neccassary? Cause i don't think it is considering 14.04 doesn't get hardware updates etc, only maintance updates and only till Q1 2019

errotan:
Extended support and updates can be bought from canonical for 14.04 until 2021 Q1.

PR #302, opened 2018-12-27
Roadmap link fix

satanch:
Fixed roadmap link in README


errotan:
The devel branch is constantly changing. My opinion is that it is not wise to point to that location.

toitzi:
Agree to @errotan, If Roadmap is "final" it will be pushed to the website, files in devel branch often change, and are not always correct/true/actual

Can close this?

neo22s:
fine by me closing

Issue #304, opened 2019-01-01
Koseven AND OR Conventions


toitzi:
According to [Koseven Conventions](https://docs.koseven.ga/guide/kohana/conventions#comparison-operations) we have to use **AND** and **OR** instead of **&&** and **||** as comparison operations. 

Is there any specific reason why?

For anyone wondering what's the diffrence, here is an example why using **&&** and **||** is considered as best practice:

```
$x = false || true;  // is ($x = (false || true)) => true
$x = false or true;  // is (($x = false) or true) => false
    
$x = true && false;  // is ($x = (true && false)) => false
$x = true and false; // is (($x = true) and false) => true
```

neo22s:
Not any specific reason, just the convention has been always in use, I think it's easier to read actually maybe I am used.

Changes the order the comparison it's done? never had any issue like that actually, but good to know.

toitzi:
@neo22s Yeah because usually comparison is not assigned to variables before it is inside the if condition. Only if you pre-assign the condition to variables this happens. [Example](http://sandbox.onlinephpfunctions.com/code/206b9a2e2b89a9da551fadba54a69939752b8aac)

piotrbaczek:
THIS IS CRITICAL. ALWAYS should && || be used.
I had this issue at my previous work. Learned never to use AND/OR operators because of PHP precedence.
http://php.net/manual/en/language.operators.precedence.php

Example:
```
<?php
$bool = true && false;
var_dump($bool); // false, that's expected

$bool = true and false;
var_dump($bool); // true, ouch!
?>
```

toitzi:
@piotrbaczek Yes that's why i posted this issue, i think we should change our conventions.
Maybe Let's Vote?

@neo22s @jstrobel @errotan @svenbw what do you think

For me +1 use && and ||

neo22s:
I am biased in this since I am really used to AND OR. We can change it internally and the convention. In my personal project it's not worth changing.

If it's such big issue +1 from me. 

We need to be super careful when replacing this.

Sebjugate:
I don't know if my vote matters, but I always use && and || in my personal code and have found it strange that Kohana uses AND/OR. 

So I would be in favor of a change in conventions to &&/||

svenbw:
For me it feels quite natural right now, precedence is important and different -or even strange- in PHP for the AND and OR operators.
[...]

PR #306, opened 2019-01-02
Userguide

toitzi:
Related to #304


piotrbaczek:
Is this for 4.0?

toitzi:
No i think we can merge it devel

PR #309, opened 2019-01-05
Updated Unittest Module

toitzi:
While discovering our unittest module overwrites white-listing files from coverage, i took time to update the code, where i recognized, there are functions inside unittest module which are not working as they supposed to be or are just very outdated/legacy code. As there are other unittests dependent on that code this was kinda critical. That's why i think writing UnitTests for our unittest module are even more important than other unittests. I will write them within the next days. Please don't merge till explicit request to do so.


toitzi:
Please review. After approval i will merge it

toitzi:
@jstrobel Maybe you got time to review? 😄

toitzi:
Merged with #322

PR #310, opened 2019-01-07
Renamed Framework to KO7

toitzi:
…  and created a Kohana module for compatibility.

See #282, #285, #289

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,27 @@
+---
+name: Bug report
+about: Create a bug report to help us improve koseven
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is, what version of koseven you are running, etc.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Additional context**
+Add any other context about the problem here.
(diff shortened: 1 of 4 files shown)


toitzi:
@neo22s @errotan @svenbw @jstrobel @piotrbaczek:

Builds Passing, Coverage remains the same (-0.01% because of adding the $compatibility variable 😄  )...looks good so far.
Tested it a bit, everything seems to work just fine. I think it is very hard to review since so much files changed, but if a few of you have bigger test projects (like @neo22s), it would be great to test it with those and check if errors occur.

Please NOTE:
This will only work if you merge correctly your `index.php` and `bootstrap.php`.

 Inside your `index.php` : 
Before the line `require APPPATH.'bootstrap'.EXT;`, `Kohana `Classes wil not be available (same as before). No need to define `KOHANA_START_TIME` and `KOHANA_START_MEMORY` as they are defined in the modules `init.php`

Inside `bootstrap.php` everywhere before `spl_autoload_register`, `Kohana` Classes won't be available. Everywhere after that you are free to use them, as they will be loaded.

Including System files directly via `require` is because of autoloader not necessary.
Only inside `index.php` and `bootstrap.php` ... so watch out for them! 

Final Note:: KO7 searches for Kohana files inside `MODPATH . 'kohana'`. So please keep the name of the module folder as 'kohana' and check if it is inside your `MODPATH` - This one was for you @neo22s 😸 as you you overwrite `MODPATH` in your yclas application.


Thanks a lot!

toitzi:
Rename from K7 to KO7

toitzi:
@jstrobel tested with a few small examples and one bit bigger application (yclas from @neo22s), everything seems to work (only method modifier was private in Encrypt (which should not be the case cause of cfs)....fixed that. What do you think, can i merge to devel to get more feedback on this and get it more tested? Merging back is always possible but changes seem pretty well working to me (even tho i will make a more detailed upgrade guide) 😄

neo22s:
You can go ahead you have my +1. The Others?

jstrobel:
+1 from me too

Issue #245, opened 2018-07-26
Support for json config files


svenbw:
At this moment all config files are PHP arrays, these arrays are difficult to update during the deployment process on the server. Typically the correct config files are copied.

If we would move from php (array based) config files to json config files the install process can be simplified.
This can be easily achieved by adding an **additional** config reader. If one switches the config reader in the bootstrap file it is possible to switch from php to json.

The structure of a combined json.config file could look like this:

```
{
    "database": {
        "default": {
            "type": "MySQLi",
            "connection": {
                ...
            }
        }
    },
    "url": {
         "trusted_hosts": [
           "localhost", "mydomain.org"
         ]
    }
}
```
It's also possible to keep the config files for the different modules separate, but the combined requires only one read.

WinterSilence:
@svenbw - change 2 strings:
~~~
class Kohana_Config_Json_Reader implements Config_File_Reader
{
    const EXT = 'json';

    public function load($group)
    {
        $config = [];
        $files = Kohana::find_file($this->_directory, $group, self::EXT, TRUE);
        if ($files) {
            foreach ($files as $file) {
                $config = Arr::merge(
                    $config, 
                    json_decode(file_get_contents($file), TRUE)
                );
            }
        }
        return $config;
    }
}
~~~

svenbw:
@WinterSilence thanks for the feedback and example. I have something similar (I've put config reader and json file reader into one class for testing; [see gist](https://gist.github.com/svenbw/d16cc8a60e8acfb17065b0590b06fd77).

I have no idea about the performance, my main issue was that I wanted to take a default config and patch it during the deployment on the server.

Take as an example the database; engine, prefix are equal for different servers; password, username, database differ; reading the default config, merge your own values, write it to the config directory and you're done.
If you start from a PHP array it's not that easy because you have to write your own php array writer (which is possible).
yclas uses default configs and does a search and replace, that's also an option.

Again: I not criticize the use of array's it's fits the Kohana/Koseven; It's easy to configure, understand and not bloated.
But looking at npm, composer, ... they all use json to store configurations. I assume because it's easy to modify trough software and is also quite readable. I opened this issue to see if others might benefit (see advantages or disadvantages) and if found useful add it to Koseven as a new feature that doesn't break things.

WinterSilence:
[...]

PR #313, opened 2019-01-13
Add support for multiple config files

toitzi:
#245  Add suport for JSON and YAML configurations.

Suports PHP Variables inside yaml and json.

Unittests need to be done, please review anyway.


toitzi:
@neo22s 

PHP:

````
<?php
return [
   'default' => [
	'type'   => 'openssl',
	'key'	 => 'someencryptionkey',
	'cipher' => Encrypt_Engine_Openssl::AES_256_CBC
   ]
]
````

JSON:

````
{
  "default" : {
    "type"   : "openssl",
    "key"    : "someencryptionkey",
    "cipher" : "<?php echo Encrypt_Engine_Openssl::AES_256_CBC; ?>"
  }
}
````

YAML:

````
default:
  type   : openssl
  key    : someencryptionkey
  cipher : <?php echo Encrypt_Engine_Openssl::AES_256_CBC; ?>
````

jstrobel:
> @jstrobel Line 63: `$config = [];`
> That's the default empty array which gets returned in case no configuration is found.
> Was that what you meant?
> 
> Reason i changed it is:
> If no configuration is specified inside APPPATH but SYSPATH, it first checks SYSPATH and get's config value and then it resetted the config to empty array cause nothing found in APPATH.

Sorry haven't just seen the "part" for the changes :) Perfect thanks.

errotan:
<?php tag in json, yaml is kinda weird and allow arbitary code to be executed (security issue?). Symfony uses this syntax: !php/const:AppBundle\Entity\BlogPost::MUM_ITEMS

neo22s:
looks good for me ;)

@errotan thanks for pointing. Weird case that can happen but better prevent it.

toitzi:
@errotan Allowed it cause we use Constants in some configs.
Why do you think this is a security issue? You can also write php code inside our .php config right now seems to be the same to me

errotan:
I'm busy studying until january 29. so can't review or test until then. Can you include unit tests?

toitzi:
@errotan added unittests

toitzi:
Think this can be merged. Feel free to revert if you disagree.

Issue #323, opened 2019-02-26
CVE-2019-8979: SQL Injection in koseven


apoleon:
Hello,

someone published a report that SQL Injection was possible in koseven/kohana when the order_by() parameter can be controlled. This was assigned CVE-2019-8979.

https://github.com/huzr2018/orderby_SQLi/tree/master/kohana

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-8979

It is not clear if the initial reporter contacted you and made you aware of the issue.

neo22s:
But this happens in a controller that does not clean the parameters I understand no?

toitzi:
@apoleon Thank you very much for providing us this information!
It is correct, that we do not escape order_by parameter.

I tested the scenarios and could not verify that an SQL injection works in the way it is stated in above documents.  I can not guarantee that there is NO way an SQL injection can't happen this way, but i can almost guarantee that the injection could not happen like in the example provided.

Althrough we understand the concern of not escaping those parameter and we will implement a check for it asap!

danielbreves:
Thanks for fixing this. Does anybody know when would be the next stable release?

jstrobel:
@danielbreves We have planned a bigger release 4.0.0 but this needs some time: https://github.com/koseven/koseven/blob/master/ROADMAP.md  .. Right now i cannot give a time for a stable 3.x release. If we have information i will let you know.

toitzi:
@danielbreves no worries...koseven is not vulnerable in current stable to the security bug above. The cve entry got corrected a while ago.

danielbreves:
@toitzi ah ok, thanks for the update. Though it seems Snyk is still reporting it as a high severity vulnerability: https://app.snyk.io/vuln/SNYK-PHP-KOSEVENKOSEVEN-173713. I'll ignore it manually for now.

toitzi:
@danielbreves Snyk gets their information from the CVE but they dont seem to update it if the CVE is updated...there is pretty much nothing we can do about that than writing to the support, but writing to all sites which do this would be a mess. My recommendation - if you use 3rd party websites - always check the CVE.

PR #324, opened 2019-02-27
Add check to order_by clause

toitzi:
See #323 

Please review and if correct merge asap.


neo22s:
ok, I agree on @errotan to send an exception if not exactly ASC or DESC. I did not think on that scenario and really easy to solve.

Thanks a lot for the job guys!

toitzi:
@errotan yes you are right, i absolutley agree with you...i updated the pr please review!

neo22s:
agree! all good ;)

svenbw:
The strtoupper should be moved from below the test to above. `asc` and `desc` are also valid options.

toitzi:
@svenbw of course you are right...where is my head right now 😸

neo22s:
@toitzi sleep deprivation!! ;) great job and thanks again

Issue #327, opened 2019-03-06
Add Support for absolute URLs in Minion


toitzi:
See: https://koseven.discourse.group/t/support-for-absolute-urls/33

PR #334, opened 2019-03-08
Add Support for absolute URLs in Minion

svenbw:
# PR Details

Create correct urls in minion scripts based on the domain name.

### Description

Add an additional function in minion module which should be called before using the urls.
Configuration can be done using the site config file or as a function argument.

### Related Issue

#327 

### How Has This Been Tested

Added unit tests.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

### Checklist

- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


modules/minion/tests/minion/TaskTest.php
```diff
@@ -67,4 +109,38 @@ public function test_convert_class_to_task($expected, $class)
...
+			['http://www.example.com/welcome', 'http://www.example.com', 'welcome'],
+			['http://www.example2.com/welcome', NULL, 'welcome'],
+			['http://www.example.com:8080/welcome', 'http://www.example.com:8080', 'welcome'],
+		];
+	}
+
+	/**
+	 * Tests that a task can be converted to a class name
+	 *
+	 * @test
```
toitzi: Remove this, not needed cause function name starts with "test"

modules/minion/tests/minion/TaskTest.php
```diff
@@ -14,6 +14,48 @@
...
+	// @codingStandardsIgnoreStart
+	public function setUp() : void
+	// @codingStandardsIgnoreEnd
+	{
+		parent::setUp();
+
+		KO7::$config->load('url')->set(
+			'trusted_hosts',
+			['www\.example\.com', 'www\.example2\.com']
+		);
```
toitzi: everytime you change some runtime configuration in the setup function, please do it via the setEnviroment function, because it creates a backup for the old config and restores it on tearDown.

Use something like:

````
$this->environmentDefault = [
	'url.trusted_hosts' => ['www\.example\.com', 'www\.example2\.com']
];
````

Comment (toitzi) [reply modules/minion/tests/minion/TaskTest.php]: also use this for other config changes in your setUp function

File: modules/minion/classes/KO7/Minion/Task.php [hunk 3]
```diff
@@ -362,4 +362,26 @@ protected function _compile_task_list(array $files, $prefix = '')
...
+	{
+		if (Request::$initial === NULL)
+		{
+			$domain_name = empty($domain_name) ? Arr::get(KO7::$config->load('site'), 'minion_domain_name', '') : $domain_name;
+
+			// Add trailing slash
+			KO7::$base_url = preg_replace('~^https?://[^/]+$~', '$0/', $domain_name);
+
+			// Set HTTPS for https based urls
+			$_SERVER['HTTPS'] = (preg_match_all('#(https)://#i', KO7::$base_url, $result) === 1);
```
toitzi: Replace this with the `$request->secure()` function

modules/minion/tests/minion/TaskTest.php
svenbw: Damn, missed this one

toitzi:
Looks good to me. Handle it as hotfix and merge it both branches or feature and merge it master?

modules/minion/tests/minion/TaskTest.php
```diff
@@ -67,4 +108,37 @@ public function test_convert_class_to_task($expected, $class)
...
+	 * @covers Minion_Task::set_domain_name
+	 * @dataProvider provider_set_domain_name
+	 * @param string Expected domain url
+	 * @param string Input domain name
+	 */
+	public function test_set_domain_name($expected, $name, $uri)
+	{
+		Minion_Task::set_domain_name($name);
+		
+		$this->assertSame(
```
toitzi: Static function should not be called dynamic

svenbw: It's a call to the Minion_Task class, not using $this.

toitzi: github web highlighted the wrong lines....my comment is related to line 139

`$this->assertSame()`

should be replaced with:

`self::assertSame()`

svenbw: Test function is not a static function, neither is `assertSame()`. So I still not get the issue.

toitzi: `Minion_TaskTest`  extends from `KO7_Unittest_TestCase` which extends from `TestCase`, which extends from `Assert` -> where the `assertSame` function is defined.

Here is the defined function:

Screenshot_3

deinfed **`static`**. I think it works via `$this->assertSame` because of Magic Methods (which should be kinda avoided), but inside the PHPUnit Framework it is also called via:
`static::assertSame`, so i thought we should also use it that way. Anyway, if you still want to use `$this` feel free to merge!

svenbw: Interesting...
If you look at the documentation of php unit the [usage example](https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertSame) of `assertSame`

[This appendix](https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.static-vs-non-static-usage-of-assertion-methods) has some more details on it. It says that there is no right or wrong way.
From my perspective I would agree with you because it's a static function so it should be used as a static function.

**However**: all tests in the Koseven use the `$this` invocation, the documentation of phpunit also the dynamic call, so I would stick with that.

But again: interesting :thinking:

toitzi: > If you look at the documentation of php unit the usage example of assertSame

It's hilarious...sometimes in PHPUnit they use `$this` sometimes they use `static`

> However: all tests in the Koseven use the $this invocation, the documentation of phpunit also the dynamic call, so I would stick with that.

Hm...yes kinda agree...since i need to rewrite many unittests i think we should move to static calls, because magic methods are kinda....yeah....you know 🙈...also my IDE warns my about this every time and it triggers me a bit 😄, but maybe in the future we can vote about that, let's keep it as it as....great work by the way! 😃 

Just last clearification for me: will it be merged to devel only as a future? Or to both devel and master?

toitzi: p.s first paragraph in the new manual for version 8.0 [Link](https://phpunit.readthedocs.io/en/8.0/assertions.html?highlight=assertSame)

> A common question, especially from developers new to PHPUnit, is whether using $this->assertTrue() or self::assertTrue(), for instance, is “the right way” to invoke an assertion. The short answer is: there is no right way. And there is no wrong way, either. It is a matter of personal preference.

😺

modules/minion/tests/minion/TaskTest.php
```diff
@@ -14,6 +14,49 @@
...
+	{
+		parent::setUp();
+
+		// Keep the old request object and base_url
+    // These are changed in set_domain_name and will cause other tests to
+    // fail, so they need to be stored and restored in the tearDown method.
+		static::$initial_request = Request::$initial;
+		Request::$initial = NULL;
+
+		static::$base_url = KO7::$base_url;
```
toitzi: Sorry my bad missed that one....you can also do all static variable changes via the $enviromentDefault function, they will be resetted afterwards and you can remove your reset from the tearDown function

toitzi: Oh and sorry for beeing so critical, i dont want to annoy you it's just that i have so much work with unittests currently (cause many of them do not work as they are intented to do) i am a bit detailed to those.....

PR #342, opened 2019-03-13
Updated image module tests

svenbw:
# PR Details

Updated the tests for the image module

### Description

Most functions are covered in tests now.

### Related Issue

No related issues.

### How Has This Been Tested

Only tests where added.

### Types of changes

- [ ] Docs change / refactoring / dependency upgrade
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [X] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #367, opened 2019-07-20
Update/image tests

toitzi:
Basically #342 (Big thanks to @svenbw ), but trying to add webp support to travis...this may take some commits...please do not merge this


toitzi:
Got another idea (maybe way cleaner and way faster) to archive this....i recently made a docker image for instant koseven deployment...i think i will create a docker image for Ubuntu 16.04 and 18.04 with everything pre-installed and configured we need and then use this docker inside travis....
i will close this and keep you updated in another issue when im done

svenbw:
Thanks for your commitment!

PR #368, opened 2019-07-21
Move travis native to docker

toitzi:
# PR Details

As described [here](https://github.com/koseven/koseven/pull/367#issuecomment-513498504)
It is easier and more clean to move to a dockerfile system instead travis native

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,3 @@
 service_name: travis-ci
 coverage_clover: build/logs/clover.xml
+json_path: build/logs/coveralls-upload.json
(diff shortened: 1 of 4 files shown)


toitzi:
@jstrobel @svenbw

Test pass (150 skipped - same as before)

 -0,01% Coverage is because of LC_Ctype issues, this should be done within the tests i will fix the tests anyway so i merge this one in favor of #342 #365 . 

If there are any conerns feel free to reopen / undo. Since tests pass i think we are fine, if someone shall be added we can do this via a issue / pr.

Issue #284, opened 2018-11-21
REST Controller


neo22s:
I've developed this API_Controller some years ago:
https://github.com/yclas/yclas/blob/master/oc/classes/api/controller.php

Some controller usage examples:
https://github.com/yclas/yclas/tree/master/oc/classes/controller/api

Authentication: In our case we create a token per user:
https://github.com/yclas/yclas/blob/master/oc/classes/auth/oc.php#L188:

Route that could be added as default:
https://github.com/yclas/yclas/blob/master/oc/config/routes.php#L264

We will need to review all code if works out of the box and make a good proper documentation with tons of examples.

Not sure if this is something we should have out of the box or in a module, but a lightweight framework as Koseven should be used by many as backend for their apps and currently it is not easy to do it and I believe it's due to the lack of such a "basic" thing ;)

Sebjugate:
Cool. I agree that this is something I suspect a lot of apps could use or have already implemented in their own way. I have implemented something similar for my own apps.

If it is to be included, I would favor a module.

toitzi:
Look at my extension: https://github.com/toitzi/koseven-rest

It's pretty much that..isn't it?

toitzi:
@neo22s I agree. REST functions is a must have for me...especially when creating APIs

I will update the docs.

toitzi:
@neo22s ok, should i do a PR?

neo22s:
can you add tests please?

toitzi:
yes

piotrbaczek:
I can look into this this weekend @toitzi if you don't have time.
Let's create Controller_Api_Base that will extend from Controller

toitzi:
@piotrbaczek would be great thanks!

toitzi:
@WinterSilence @piotrbaczek @neo22s  I created a PR (#374) regarding this (not final, just a quick implementation),
i took some parts of your code @WinterSilence and some of the original one and put it together, i also added a REST_Response class which allows passing arrays to the body (very common for REST Api's)....please let me think if you think this is the right direction...here are 2 examples of how to use it:

### For both examples
Simply change inside `Controller_Welcome`:
`extends Controller` to `extends Controller_REST`.

### 1
That's already it!

Now the output should be:
`{"body":"hello, world!"}` (with `application/json` as header)

### 2
You can also change the line `$this->response->body('hello, world!');` to something like:
`$this->response->body(['id' => 1, 'name' => 'Koseven']);`

Now the output should be:
[...]

PR #374, opened 2019-08-01
Implement a REST_Controller

toitzi:
See #284


toitzi:
@jstrobel @neo22s @WinterSilence @svenbw @piotrbaczek 
This is how i would implement it, and i really would like to here your feedback and if it is ok this way, if you approve i will write the documentation + unittests and merge it devel afterwards.

If you know my original module these are the main differences are:

1. Format methods are moved into seperate format classes (This allows easy implementation of custom formatter and follow KO7 style)

2. Like @WinterSilence suggested: Everything is moved to built-in output and `before` and `after` classes which makes it much more convenient and clean.

3. Even existing controller can be converted into "REST" controller, just extend `Controller_REST` instead of `Controller`, everything should work like it did before - except your output is REST (default is `JSON`)

of course it is updated to PHP7.3, new conventions and new Class names, legacy code got removed

toitzi:
p.s i know code formatting is a bit messy...but i will ofc fix it within my next commits

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,198 @@
...
+			'_status'   => $response->status(),
+			'_body'     => $response->body(),
+			'_cookies'  => $response->cookie(),
+			'_protocol' => $response->protocol()
+		];
+
+		// Copy all http header(s)
+		foreach ($response->headers()->getArrayCopy() as $header => $value)
+		{
+			$config[$header] = $value;
```
WinterSilence: why not `$config += $response->headers()->getArrayCopy(); `?

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,198 @@
...
+
+		// Format body
+        $body = $this->_formatter->format($body);
+
+		// Parse and set response headers
+		$this->response->headers('content-type', $this->_content_type ?? File::mime_by_ext($this->output_format));
+		$this->response->headers('content-length', (string) strlen($body));
+
+		// Support attachment header
+		if (isset($this->_params['attachment']) && Valid::regex($this->_params['attachment'], '/^[-\pL\pN_, ]++$/uD'))
```
WinterSilence: it's not for controller as half this code

system/classes/KO7/REST/Format/XML.php
```diff
@@ -0,0 +1,41 @@
...
+    public function format(array $body): string
+    {
+        // Check if php-xml is loaded
+        if ( ! extension_loaded('xml'))
+        {
+            throw new REST_Exception('PHP XML Module not loaded.');
+        }
+
+        // Create new XML Element
+        $xml = new SimpleXMLElement('<root/>');
```
WinterSilence: use power of PHP 7.3

WinterSilence:
@toitzi I can't use this library anymore and staying here as spectator

system/classes/KO7/Controller/REST.php
toitzi: Yes, Good one thanks!

system/classes/KO7/Controller/REST.php
toitzi: What do you mean?

system/classes/KO7/REST/Format/XML.php
toitzi: I did not find another option to do this. SimpleXMLElement is the php-xml extension....don't know/think that there is a native function

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,198 @@
...
+	 * Default Output (only if allowed by configuration)
+	 * @var string
+	 */
+	public static $default_output = 'json';
+
+	/**
+	 * The output format to be used (JSON, XML etc.).
+	 * @var string
+	 */
+	public $output_format;
```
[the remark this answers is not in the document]
toitzi: Content-Type is optional and evaluated form output_format, also converting mime to extension would be a bit long, or what exactly do you mean?

toitzi:
> I can't use this library anymore and staying here as spectator

Thanks a lot for staying as spectator and your feedback!

system/classes/KO7/Controller/REST.php
WinterSilence: logic in controller, not SOLID

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,198 @@
...
+				'body' => $body
+			];
+		}
+
+		// Format body
+        $body = $this->_formatter->format($body);
+
+		// Parse and set response headers
+		$this->response->headers('content-type', $this->_content_type ?? File::mime_by_ext($this->output_format));
+		$this->response->headers('content-length', (string) strlen($body));
```
WinterSilence: https://koseven.ga/documentation/api/Response#render

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,198 @@
...
+     *
+     * @throws HTTP_Exception
+	 */
+	public function before() : void
+	{
+		// Parent call
+		parent::before();
+
+        // Try fetching method from HTTP_X_HTTP_METHOD_OVERRIDE, otherwise use the one sent with the request
+        $this->request->method($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE'] ?? $this->request->method());
```
WinterSilence: **always** use `filter_var()` for global variables

system/classes/KO7/REST/Format/XML.php
WinterSilence: `$xml = new SimpleXMLElement('<?xml version="1.0" encoding="'.KO7::$charset.'"?><root/>', LIBXML_BIGLINES | LIBXML_COMPACT | LIBXML_NOCDATA | LIBXML_NOENT);`
also u can add  `| LIBXML_NOERROR | LIBXML_NOWARNING` to options and check `libxml_get_errors()` for throw exception

system/classes/KO7/Controller/REST.php
toitzi: Thanks removed the content-length

system/classes/KO7/REST/Format/XML.php
toitzi: Thanks, done and added error check / throw exception

system/classes/KO7/Controller/REST.php
toitzi: Added inline check instead

system/classes/KO7/REST/Format/XML.php
```diff
@@ -0,0 +1,41 @@
...
+
+        // Add Child foreach body element
+        $xml = $xml->addChild('data');
+        array_walk_recursive($data, static function($value, $key) use ($xml, &$result) {
```
WinterSilence: 1. $result` not used
2. wrong xml, add tests

WinterSilence:
~~~
$data = ['a' => ['a2' => ['a3' => 1]], 'b' => 1];
~~~
$xml->asXml():
~~~
<data><a3>1</a3><b>1</b></data>
~~~

system/classes/KO7/Controller/REST.php
toitzi: Agree, i will take a closer look into this and check if i can move those functions where they belong

toitzi: Moved stuff belonging to the response class, into there,...will move the ones belonging to the request class within this week than we should be good to go

system/classes/KO7/REST/Format/XML.php
toitzi: Agrre, thanks, will fix this, i write tests when code is done, therefore i do not waste time rewriting them everytime

WinterSilence:
~~~
/**
 * Convert array to xml.
 * 
 * @param array $data
 * @param string|SimpleXMLElement $root
 * @param string $charset
 * @return SimpleXMLElement
 */
public function arrayToXml(array $data, $root, $charset = null)
{ 
    if (is_string($root)) {
        $root = new SimpleXMLElement(
            '<?xml version="1.0" encoding="'.($charset ?: KO7::$charset).'"?><'.$root.'/>',
            LIBXML_BIGLINES | LIBXML_NOCDATA | LIBXML_NOENT
        );
    }
    
    foreach ($data as $name => $value) {
        if (substr($name, 0, 1) == '@') {
            $root->addAttribute($name, $value);
        } else {
            if (is_array($value)) {
                $this->arrayToXml($value, $root->addChild($name), $charset);
            } else {
                $root->addChild($name, $value);
            }
        }
    }
    
    return $root;
}
~~~

toitzi: Fixed, also fixed the content-type

system/classes/KO7/Controller/REST.php
toitzi: Moved functions and operations where it was possible to there files (Response and Request), moved some functions to already implemented fuctions and removed unneccessary properties from classes....

system/classes/KO7/REST/Format/JSON.php
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * Format function
+     *
+     * @param array $body Body to format
+     *
+     * @return string
+     */
+    public function format(array $body) : string
+    {
+        return json_encode($body);
```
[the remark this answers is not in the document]
toitzi: Thanks, done...also added correct exception handling

system/classes/KO7/REST/Format/JSON.php
```diff
@@ -0,0 +1,58 @@
...
+                |JSON_NUMERIC_CHECK
+                |JSON_PRESERVE_ZERO_FRACTION
+                |JSON_UNESCAPED_UNICODE
+            );
+        }
+        catch(Exception $e)
+        {
+            $message = 'Unknown JSON error.';
+
+            switch(json_last_error()) {
```
WinterSilence: `json_last_error()` doesnt work at `JSON_THROW_ON_ERROR` replace to 
`throw new REST_Exception($e->getMessage(), NULL, $e->getCode(), $e);`

system/classes/KO7/REST/Format/XML.php
```diff
@@ -0,0 +1,113 @@
...
+     * @param null $xml           Current XML stack
+     *
+     * @return mixed
+     */
+    private function array_to_xml(array $array, $rootElement = NULL, $xml = NULL)
+    {
+        // If there is no Root Element then insert root
+        if ($xml === NULL)
+        {
+            $xml = new SimpleXMLElement($rootElement ?? '<?xml version="1.0" encoding="'.KO7::$charset.'"?><root/>', LIBXML_COMPACT);
```
WinterSilence: 1. u cant catch error as u want because remove `LIBXML_NOWARNING | LIBXML_NOERROR`
2. `LIBXML_BIGLINES` too important
3. dont set methods as `private` - hard to extending this classes

system/classes/KO7/REST/Format/JSON.php
toitzi: Yeah you are right i misread that....i shouldn't work late night 😄 ...fixed it

system/classes/KO7/REST/Format/XML.php
toitzi: 1. why?
2. Added
3. Moved to protected

toitzi: Sure..got the first one...no worries....fixing it with libxml_use_internal_errors(true)

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,134 @@
...
+     * Automatically executed before the controller action.
+     * Evaluate Request (method, action, parameter, format)
+     */
+    public function before() : void
+    {
+        // Parent call
+        parent::before();
+
+        // Inject output format if not set by route
+        $this->request->inject_param('format', $this->request->param('format') ?: static::$default_output);
```
WinterSilence: 1. $this->request->param() have second param `$default`
2. format can contains in param['format'], query['format'] or header['accept']
3. Method `inject_param` is wrong way, dont rewrite request data. Why u want contains format as request param?

~~~
// replace static $default_output to
protected $format = 'json';
...
if ($this->request->param('format')) {
 $this->format = $this->request->param('format');
} elseif ($this->request->query('format')) {
 $this->format = $this->request->query('format');
} elseif ($this->request->accept_type()) {
  list(, $this->format) = explode('/', $this->request->accept_type()[0]);
}
~~~

system/classes/KO7/Controller/REST.php
```diff
@@ -0,0 +1,134 @@
...
+     * @var string
+     */
+    public static $default_output = 'json';
+
+    /**
+     * REST types
+     *
+     * @var array
+     */
+    protected static $_action_map = [
```
WinterSilence: why is static property?

system/classes/KO7/REST/Format/XML.php
```diff
@@ -0,0 +1,119 @@
...
+     * @param array $array        Array to convert
+     * @param mixed $rootElement  Root element of entry
+     * @param null $xml           Current XML stack
+     *
+     * @return mixed
+     */
+    protected function array_to_xml(array $array, $rootElement = NULL, $xml = NULL)
+    {
+        // Use internal error handling
+        libxml_use_internal_errors(true);
```
WinterSilence: Pls, if u dont understand then ask me
As I say, need add `LIBXML_NOERROR | LIBXML_NOWARNING`. You can ask me - why not libxml_use_internal_errors? Because u CHANGE GLOBAL SETTING

toitzi: I've read about some issues using `LIBXML_NOERROR`, very common in php therefore i used the function, i understand the issue of changing a global setting, that's why i am backing up the old setting and restoring it afterwards now.

system/classes/KO7/Controller/REST.php
toitzi: Probably because of a plugin from my IDE which converts "heavy" initialized variables to static ones..... after some research i have not really found a reason for this and asked (because i am curios) the developer why....maybe it is considered some kind of best practice...but anyway i am not really a fan of static stuff, i agree and moved it to non static

system/classes/KO7/Controller/REST.php
toitzi: 1. Yes but does not allow to inject one
2. Not sure if we should let the format option decide via accept header.... i think a predefined one (either via routes or as default) is enough. Your code will not work if the user adds custom driver
3. For me it is because the format is part of the request (User requests data in xx format)....i am also a bit unsure if injecting is the right thing....maybe i find a better way to archive this.

toitzi: Removed the `inject_param` function, added a `$_format` property to request which can be recived or set via `format` method (like other variables in koseven).
Now the requested format can be found there and be easy accessed within the application, because as said the "**requested** format" belongs to me in the "**request**" class

For now the format is received via route parameter, i will check if setting via accept header will be implemented....

WinterSilence:
@toitzi what's problem with `LIBXML_NOERROR | LIBXML_NOWARNING`? can u show error examples?

WinterSilence:
@toitzi also move all stuff as default_format from controller to KO7_REST_Format

toitzi:
> also move all stuff as default_format from controller to KO7_REST_Format

Done! 👍 

> can u show error examples?

here is an example: https://stackoverflow.com/a/41845049 ... seems like they fixed it, but just to be sure i don't mind using the function

WinterSilence:
@toitzi where u use DOMDocument?

toitzi:
> where u use DOMDocument?

Nowhere

WinterSilence:
@toitzi than see u link again and fix as i say...

toitzi:
> than see u link again and fix as i say...

i know that it is DOM Document, but it proved to not work sometimes, and to make sure i use the function - i see no problem doing this and therefore will keep it that way

WinterSilence:
@toitzi i use it many years and dont have promblems, trust me. libxml_use_internal_errors creates more problem and u dont know how use it - dont see •libxml_clear_errors before and u code show old errors all fine

toitzi:
> creates more problem
i don't see why

> libxml_clear_errors before
Should not be neccesary but yes we can do it in case the user forgot to in his application

> u dont know how use it
The way it is implemented now is pretty much the same as in the PHP-Documentation - as said will keep it that way

Anyway thanks for your detailed feedback and review it helps a lot!

toitzi:
> facepalm, dont call me more if dont want listen

I listen and changed many things according to your feedback, but don't always excpect that your way is the only right one and best one and always has to happen like you want.

I appreciate your feedback, but if you can not accept that not everything is implemented like you want i understand and will not call you for further. Thanks

WinterSilence:
@toitzi 
> The way it is implemented now is pretty much the same as in the PHP-Documentation - as said will keep it that way

this is wiki docs: sometimes text wrong, sometimes old. i fix some errors too. see how many updates wait accept.
Before you show me links with talk about bug and after this talk about PHP-Documentation -  its unlogic: you dont know how work libxml_use_internal_errors - good or bad, but u dont know how work LIBXML_* options too

toitzi:
@WinterSilence It works now, let's keep it

WinterSilence:
@toitzi u dont know it

toitzi:
@jstrobel For you to keep track of the current state:

I am currently writing documentation for it. Afterwards i will unittest it.

I'll notify you and merge when done.

toitzi:
@jstrobel Added Documentation, will unittest now and then merge

system/classes/KO7/Request.php
```diff
@@ -955,8 +986,14 @@ public function method($method = NULL)
 			return $this->_method;
 		}
 
+		// Method is always uppercase
+		$method = strtoupper($method);
+
+		// Allow overriding method
+		$override = isset($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']);
```
WinterSilence: use `$request->headers()`

system/classes/KO7/REST/Format/XML.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Convert Array to an xml element
+     *
+     * @param array $array        Array to convert
+     * @param mixed $rootElement  Root element of entry
+     * @param null $xml           Current XML stack
+     *
+     * @return mixed
+     */
+    protected function array_to_xml(array $array, $rootElement = NULL, $xml = NULL)
```
WinterSilence: `array/is_array` deprecated, use `iterable/is_iterable`

system/classes/KO7/REST/Format/HTML.php
```diff
@@ -0,0 +1,51 @@
...
+class KO7_REST_Format_HTML extends REST_Format {
+
+    /**
+     * Format function
+     *
+     * @throws REST_Exception
+     *
+     * @return string
+     */
+    public function format() : string
```
WinterSilence:
~~~
$path = array_filter([
$this->_request->directory(), 
$this->_request->controller(), 
$this->_request->action(),
]);
$path = strtolower(implode(DIRECTORY_SEPARATOR, $path));
if (! KO7::find_file('views', $path)) {
   $path = dirname($path);
}
return View:factory($path,  $this->_body)->render();

system/classes/KO7/REST/Format.php
```diff
@@ -0,0 +1,119 @@
...
+        // Check if client extends Request_Client_External
+        if ( ! $formatter instanceof REST_Format)
+        {
+            throw new REST_Exception(':formatter is not a valid REST formatter.', [
+                ':formatter' => get_class($formatter)
+            ]);
+        }
+
+        // Set response content type by format used
+        $response->headers('Content-Type', File::mime_by_ext($request->param('format')));
```
WinterSilence:
~~~
$type = File::mime_by_ext($request->param('format', KO7::$content_type));
$response->headers('Content-Type', $type);
~~~

system/classes/KO7/Request.php
toitzi: Corrected!

system/classes/KO7/Request.php
```diff
@@ -990,7 +990,7 @@ public function method($method = NULL)
 		$method = strtoupper($method);
 
 		// Allow overriding method
-		$override = isset($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']);
+		$override = $this->headers('HTTP_X_HTTP_METHOD_OVERRIDE');
```
WinterSilence: i think name changed to `X-Http-Method-Override`

toitzi: yes it changed

system/classes/KO7/REST/Format/XML.php
toitzi: Good idea

system/classes/KO7/Request.php
WinterSilence: @toitzi see https://koseven.ga/documentation/api/HTTP#request_headers

system/classes/KO7/REST/Format/HTML.php
toitzi: Nice way, i changed it a bit to fit into the correct order

PR #383, opened 2019-09-08
Fix bugs and code style in KO7_Config_File_Reader

WinterSilence:
1. Why used `@codeCoverageIgnore` for test env? Why tests skipped? - Undefined `self::$_cache` 100% generate notice.
2. Where demo json/yaml configs?
3. `self::load()` so fat and unextandable - need separate loading by file type (3 providers/methods)

### DIFF
--- a/system/classes/KO7/Config/File/Reader.php
+++ b/system/classes/KO7/Config/File/Reader.php
@@ -6,9 +6,9 @@
  * @package    KO7
  * @category   Configuration
  *
- * @copyright  (c) 2007-2016  Kohana Team
- * @copyright  (c) since 2016 Koseven Team
- * @license    https://koseven.ga/LICENSE
+ * @copyright  (c) 2007-2016 Kohana team
+ * @copyright  (c) 2016-2019 Koseven team
+ * @license    https://koseven.ga/LICENSE.MD
  */
 class KO7_Config_File_Reader implements KO7_Config_Reader {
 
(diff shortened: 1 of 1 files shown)


system/classes/KO7/Config/File/Reader.php
```diff
@@ -19,82 +19,85 @@ class KO7_Config_File_Reader implements KO7_Config_Reader {
...
+			{
 				$value = KO7::load($path);
-			} elseif (file_exists($path = $file.'.json')) {
-				$value = json_decode($this->read_from_ob($path), true);
-			} elseif (file_exists($path = $file.'.yaml')) {
-				if ( ! extension_loaded('yaml')) {
+			}
+			elseif (file_exists($path = $file.'.json'))
+			{
+				$value = json_decode($this->read_from_ob($path), TRUR);
```
toitzi: Bug, please change to "TRUE"?

system/classes/KO7/Config/File/Reader.php
```diff
@@ -19,82 +19,85 @@ class KO7_Config_File_Reader implements KO7_Config_Reader {
...
 	 *
-	 * @return  array   Configuration
+	 * @return  array  Configuration
 	 * @throws KO7_Exception
 	 */
 	public function load($group) : array
 	{
 		// Check caches and start Profiling
-		if (KO7::$caching && isset(self::$_cache[$group]))
+		if (KO7::$caching AND isset(static::$_cache[$group]))
```
toitzi: Please use &&

toitzi:
Thanks for your PR.

> Where demo json/yaml configs?
There are no, would be cool if you could add those into the documentation, i think there are some in the original PR

> self::load() so fat and unextandable
I think it is not necessary to make it possible adding more drivers then those, but i am not against it, i would appriciate it if you got time to implement, if not so i think we can keep it..

>  Why tests skipped?
If test got skipped, maybe check if you have installed yaml and json correctly

>  Why used @codeCoverageIgnore for test env?
Testing the `static::$cache` does not seem neccesary, also it is protected, meaning to test if it works we would have to do some Mocking i guess....but that was just a quick first look

system/classes/KO7/Config/File/Reader.php
WinterSilence: it's wrong kohana code style

toitzi: no we switched to &&

PR #420, opened 2020-09-02
Feature/cookie same site

### LINKED ISSUE #418: Missing Cookie Attribute "SameSite" = None | Strict | Lax
Look here https://www.chromium.org/updates/same-site

This new attribute is missing in https://github.com/koseven/koseven/blob/master/system/classes/Kohana/Cookie.php

PHP https://www.php.net/manual/en/function.setcookie.php
setcookie('key', 'value', ['samesite' => 'None', 'secure' => TRUE]);

piotrbaczek:
# PR Details

This PR implements Cookie property "SameSite".

### Description

1. Added 2 classes: Kohana_Cookie_Properties with dictionary related to general cookie properties and Kohana_Cookie_Samesite with settings related to SameSite property.

2. Class Kohana_Cookie method setcookie got extended with new property $samesite. This method now uses alternative call to PHP native setcookie function, with data passed as an associative array.

### Related Issue

None

Please link to the issue here

https://github.com/koseven/koseven/issues/418

### How Has This Been Tested

This feature has been tested by automatic unit test - adding required cookie parameter to already existing unittest.
Secondary method of testing was done manually by adding following line in Controller_Welcome
``
Cookie::set('aaa','bbb');
``
and manually verifying existence of Cookie in Chrome Developer's Console's Application\Cookies tab.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Checklist

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


toitzi:
@piotrbaczek Thank you i will look into this pr, i will probably open a new one and merge this devel since it is a new feature.

piotrbaczek:
I'd like to know if this can be merged and if not then why.

toitzi:
@piotrbaczek  Unfortunately not. As far as i can tell it is a new feature and adds fuctionality, this is the reason why it needs to be merged into devel (next feature release) not into master (current feature release). Usually only bug and hotfixes in master (or documentation fixes and that) small features (probably like this one) may also be merged in master but i need time to check it.

Daijobou:
If this is for the next major version, which should support php 8.1, shouldn't it use maybe **enums** instead of two extra classes [Kohana_Cookie_Samesite, Kohana_Cookie_Properties]? See https://stitcher.io/blog/php-enums
I'm not that big fan for two extra files because few constants. And I don't think you have to explain every constant in such detail. The following URL as source would be sufficient: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

P.S. In comment above "**protected static function _setcookie(**" there is parameter "$samesite" missing.

piotrbaczek:
Closing because of inactivity.

Issue #429, opened 2020-10-20
Memcache regression with 3.X functionaltiy regarding `instant_death` on falure


Chipcius:
**Describe the bug**
Instant death of memcache servers on failure was configured per server and all documentation says it is per server but someone has changed the code to look for a global for that config and without any exception handing

**To Reproduce**
1. Have a system that is using the Memcace module as cache
2. Call the cache while the Memcache server is experiencing problems
3. You'll get unexpected behavour
  - If the config (that differs from documentation) is not set. You will get an "Undefined index: instant_death" error
  - If you have the config and the individual server configs from the documentation. The server config will not work

**Expected behavior**
- As this is not a major version there should be no backwards incomparable changes. Or as few as possible
- Documentation should reflect actual behaviour

**Screenshots**
![Selection_124](https://user-images.githubusercontent.com/1033039/96602960-1985b000-12e3-11eb-9295-a5ccbf7107c0.png)

PR #430, opened 2020-10-20
Fixed regression in memcache service `instant_death` on failure

Chipcius:
# PR Details

In this PR I am re-introducing functionality that had been removed when the repository was initialized.

### Description

When Koseven was cloned in it cointained a change in codebase of where we were calling the the `instant_death` variable when memcached fails.

This looks like a start of a change that was never finished. All documentation, other code and references point to the prior functionality.

### Changes
- Fixed regression in behaviour when disabling memcache server on failure per server
- Fixed calling an undocumented member of memcache configuration to prevent failure
- Fixed some indentation in a comment that has tabs in stead of spaces

### Related Issue
https://github.com/koseven/koseven/issues/429

### How Has This Been Tested
By running this on our production system without it gobbling errors every hour

### Types of changes

- [X] Docs change / refactoring / dependency upgrade
- [X] Bug fix (non-breaking change which fixes an issue)

### Checklist

- [X] I have updated the documentation accordingly.

### DIFF
--- a/modules/cache/classes/Kohana/Cache/Memcache.php
+++ b/modules/cache/classes/Kohana/Cache/Memcache.php
@@ -289,7 +289,7 @@ class Kohana_Cache_Memcache extends Cache implements Cache_Arithmetic {
 	{
 		// This is for compatability with anyone using the config
 		// variable from Koseven initail update
-		if ($this->config('instant_death') === false)
+		if ($this->config('instant_death') === FALSE)
 			return;
 
 		// Setup non-existent host
--- a/modules/cache/classes/Kohana/Cache/Memcache.php
+++ b/modules/cache/classes/Kohana/Cache/Memcache.php
@@ -24,7 +24,7 @@
  *                              'timeout'          => 1,
  *                              'retry_interval'   => 15,
  *                              'status'           => TRUE,
- *				'instant_death'	   => TRUE,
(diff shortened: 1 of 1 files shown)


modules/cache/classes/Kohana/Cache/Memcache.php
```diff
@@ -287,7 +287,9 @@ public function delete_all()
 	 */
 	public function _failed_request($hostname, $port)
 	{
-		if ( ! $this->_config['instant_death'])
+		// This is for compatability with anyone using the config
+		// variable from Koseven initail update
+		if ($this->config('instant_death') === FALSE)
```
Chipcius, line 292: Updated this to use the built in config function

This will return `NULL` when the value is not set so we will pass through as the inner `instant_death` will be the deciding factor then

image

WinterSilence:
@Chipcius send to `devel` branch, please

Chipcius:
@WinterSilence `devel` no longer contains the `memcache` cache file and that branch has diverged from the kohana base as much as to be a major version increase.

This change as it is non breaking in any case and solves backwards incompatibilities that have been introduced by accident would count as a minor fix. There is no new feature introduced in this change.

Chipcius:
Not gotten a feedback for a while so I am giving up on this

PR #432, opened 2020-10-21
ErrorException [ Deprecated ]:

semenov-anton-a:
ErrorException [ Deprecated ]: array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead

# PR Details

Provide a general summary of your changes in the Title above

### Description

ErrorException [ Deprecated ]: array_key_exists(): 
Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead

### How Has This Been Tested

Minion error  then uncommet
Kohana::init  [
cache
]

### DIFF
--- a/system/classes/Kohana/Arr.php
+++ b/system/classes/Kohana/Arr.php
@@ -101,7 +101,7 @@ class Kohana_Arr {
 		}
 		else
 		{
-			if (array_key_exists($path, $array))
+			if (isset($path, $array))
 			{
 				// No need to do extra processing
 				return $array[$path];


system/classes/Kohana/Arr.php
```diff
@@ -101,7 +101,7 @@ public static function path($array, $path, $default = NULL, $delimiter = NULL)
 		}
 		else
 		{
-			if (array_key_exists($path, $array))
+			if (isset($path, $array))
```
Chipcius, line 104: This is not going to work as expected. 

The easier solution here is not to pass an object into the `$array` value as this expected the value to be an array. Not an object

See: https://www.php.net/isset

WinterSilence:
it's already fixed in [devel](https://github.com/koseven/koseven/blob/devel/system/classes/KO7/Arr.php)

PR #463, opened 2022-02-02
PHP 8.1 compatibility

bwdutton:
# PR Details

Make compatible with PHP 8.1

### Description

Mostly comments to loosen the type checks. The syntax for serialization in the ORM module had to be updated to be compatible with the new magic methods.

### Related Issue

This project only accepts pull requests related to open issues

If suggesting a new feature or change, please discuss it in an issue first

If fixing a bug, there should be an issue describing it with steps to reproduce

Please link to the issue here

### How Has This Been Tested

Tested in my app with PHP 8 and 8.1. I get the feeling there may need to be some more ReturnTypeWillChange additions as I didn't test all of the code.

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

### Checklist

### DIFF
--- a/modules/database/classes/Kohana/Database/Result.php
+++ b/modules/database/classes/Kohana/Database/Result.php
@@ -206,6 +206,7 @@ abstract class Kohana_Database_Result implements Countable, Iterator, SeekableIt
 	 *
 	 * @return  integer
 	 */
+	#[\ReturnTypeWillChange]
 	public function count()
 	{
 		return $this->_total_rows;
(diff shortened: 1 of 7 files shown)


Azuka (approved):

Daijobou:
Nice work and thanks for your effort. 👍 However, in the long run, wouldn't it make more sense to give the functions the correct types for parameters and returns, instead only suppress warnings and errors? I'm not that expert to do it. ;)

But probably the project is no longer maintained, especially because it is not future-proof with old php 7.x code, which is deprecated. PHP 8.1 End-of_life is in 5 months. https://www.php.net/supported-versions.php

Issue #466, opened 2022-02-09
Loading values from config files with dot notation gives NULL


ajuszczyk:
I have upgraded an old kohana system to koseven and once switched the php to version 7.4.2 suddenly all values from config files like:
`Kohana::$config->load('language.abbr')` etc give back null.
When I load `Kohana::$config->load('language')` there is data returned.
Also `Kohana::$config->load('language')->abbr` and `Kohana::$config->load('language')->get('abbr')` both work fine.

When switched back to kohana and php 5.4 values are loaded fine.

I just can't figure out why the dot notation doesn't work in koseven, any help will be much appreciated.

timhj:
Is your language config file returning an Array or StdClass? If an stdclass/object, return an array.

The config loader uses Arr::Path, which would have started failing on objects in PHP 7.4.x. As a quick fix to get things working with 7.4, the Arr::path behaviour was changed to work only with arrays (and not traversable objects). It now returns NULL if an object is passed so could be the cause of your issue. Sorry if that's your issue, it was my PR and a really quick fix to make the framework load with PHP 7.4.

ajuszczyk:
In addition this is what I'm getting back with Kohana::$config->load('language')

```
object(Config_Group)[6]
  protected '_parent_instance' => 
    object(Config)[3]
      protected '_sources' => 
        array (size=1)
          0 => 
            object(Config_File)[5]
              ...
      protected '_groups' => 
        array (size=1)
          'language' => 
            object(Config_Group)[6]
              ...
  protected '_group_name' => string 'language' (length=8)
  private 'storage' (ArrayObject) => 
    array (size=5)
      'language' => string 'english' (length=7)
      'language_abbr' => string 'en-gb' (length=5)
      'lang_uri_abbr' => 
        array (size=3)
          'pl-pl' => string 'polish' (length=6)
          'en-gb' => string 'english' (length=7)
          'en-us' => string 'us_english' (length=10)
      'lang_ignore' => string 'xx' (length=2)
      'lang_desc' => 
        array (size=3)
          'en-us' => string 'US English version' (length=18)
          'en-gb' => string 'English version' (length=15)
          'pl-pl' => string 'Polish version' (length=14)
```

timhj:
@ajuszczyk  - I mentioned two fixes, did you apply this one: https://github.com/koseven/koseven/pull/467/files ? 

It looks like it should solve your issue as you're getting individual values from the config array, which would not have worked.

[...]

PR #468, opened 2022-02-11
Fix for Issue #466 - Arr::path to check for array and/or traversable …

ajuszczyk:
Use Arr::is_array() to return correct bool value for checking against an array and/or traversable object.

# PR Details

Using just is_array() was returning false in case of traversable object. This broke the dot notation when getting back the config files values.

### Description

Used helper function Arr::is_array() that check for an array or traversable object.

### Related Issue

It's provided as a fix for [https://github.com/https://github.com/koseven/koseven/issues/466](Issue https://github.com/koseven/koseven/issues/466)

### How Has This Been Tested

Small change, quick smoke test

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.

### DIFF
--- a/system/classes/Kohana/Arr.php
+++ b/system/classes/Kohana/Arr.php
@@ -88,7 +88,7 @@ class Kohana_Arr {
 	 */
 	public static function path($array, $path, $default = NULL, $delimiter = NULL)
 	{
-		if ( ! is_array($array))
+		if ( ! Arr::is_array($array))
 		{
 			// This is not an array!
 			return $default;


timhj:
There was something deeper in Arr::Path that failed when traversing objects from PHP 7.4+ but can't remember exactly. It was the reason for only allowing arrays (is_array() vs. Arr::is_array()) to be passed to the function.

PR #496, opened 2022-07-30
Fix uniqid() arguments

atainex:
Fixing uniqid arguments for PHP 8


toitzi:
Tanks a lot, looks good is merged!

PR #503, opened 2023-04-05
Valid::numeric local independent

### LINKED ISSUE #502: [PHP 8.0] Valid::numeric should be locale independent
**Describe the bug**
When using the Valid::numeric method on a float in PHP >=8.0 with a locale that does not use a dot (.) as a decimal separator, a correct numeric value would not pass.

**To Reproduce**
1. Change [bootstrap.php line 33](https://github.com/koseven/koseven/blob/master/application/bootstrap.php#L33) to `setlocale(LC_ALL, 'fr_FR');`
2. Call `Valid::numeric(1.45);`

**Expected behavior**
Expected result is `true` but `false` is returned

**Additional context**
[https://php.watch/versions/8.0/float-to-string-locale-independent](https://php.watch/versions/8.0/float-to-string-locale-independent)

CasperBE:
Fixes #502

### DIFF
--- a/system/classes/Kohana/Valid.php
+++ b/system/classes/Kohana/Valid.php
@@ -433,15 +433,21 @@ class Kohana_Valid {
 	 * Checks whether a string is a valid number (negative and decimal numbers allowed).
 	 *
 	 * Uses {@link http://www.php.net/manual/en/function.localeconv.php locale conversion}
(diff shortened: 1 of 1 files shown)


toitzi:
LGTM thanks!

